PR #3, opened 2020-11-19
Fix boolean values

mietcls:
Not sure if the @disabled="{{true}}" should have been converted to @disabled={{true}}; which is why I am sending a pull request.


aatauil (commented):
The changes dont have an immediate effect since {{true}} & "true" both equal the boolean True. But {{true}} is more, call it "semantically" correct and prefered :)

aatauil:
stale PR. Closing as this doesnt seem to give any issues

PR #6, opened 2020-12-21
Upgrade app to 3.20

aatauil:
Everything looks fine, Besides a couple of this keywords missing I did not find syntax that needed to be upgraded


claire-lovisa:
Waiting for Loket's upgrate before merging to ensure it still integrates properly

HugaertsDries:
@aatauil I've been trying to link this branch off the plugin in to `frontend-meldingsplichtige-api`master (merged the upgrade to 3.20), but no luck. It seems to be silently failing/breaking the render. I've not yet been able to isolate the problem myself.

Node `v12.20.1`
NPM `6.14.10`

HugaertsDries:
> @aatauil I've been trying to link this branch off the plugin in to `frontend-meldingsplichtige-api`master (merged the upgrade to 3.20), but no luck. It seems to be silently failing/breaking the render. I've not yet been able to isolate the problem myself.

So after some digging I uncovered the issue. Could be very specific to the consuming app `frontend-meldingsplichtige-api`, but could reoccur if other consuming apps if the same mistake or assumption was made . 

What happend is that `ember-moment` got moved from `dependencies` to `devDependencies` within the plugin. The consuming app relied on the plugin to supply this dependency, to make use off the helper moment-format. Because the dependency was not there anymore, it couldn't find the helper. Hence, it broke.

**All in all, the plugin itself seems to work fine**.

HugaertsDries (approved):

PR #8, opened 2021-01-25
Feature/radio buttons

aatauil:
It is similar to conceptSchemeSelector so most of the code comes from that component


addon/components/rdf-input-fields/concept-scheme-radio-buttons/edit.js
```diff
@@ -0,0 +1,71 @@
...
+import rdflib from 'browser-rdflib';
+
+
+function byLabel(a, b) {
+  const textA = a.label.toUpperCase();
+  const textB = b.label.toUpperCase();
+  return (textA < textB) ? -1 : (textA > textB) ? 1 : 0;
+}
+
+export default class RdfInputFieldsconceptSchemeRadioButtonsEditComponent extends SimpleInputFieldComponent {
```
HugaertsDries, line 16: Kudos for extending from `SimpleInputFieldComponent`! 
What made you do so? I'm assuming to make use of `nodeValue`?

addon/components/rdf-input-fields/concept-scheme-radio-buttons/edit.js
```diff
@@ -0,0 +1,71 @@
...
+  inputId = 'conceptSchemeRadioButtons-' + guidFor(this);
+
```
HugaertsDries, line 18: Noticed that you removed the `@tracked` properties? Don't they need to be in a `get` method for them not to be explicitly defined? Or does an `@action` tag also make them tracked?

addon/components/rdf-input-fields/concept-scheme-radio-buttons/show.js
```diff
@@ -0,0 +1,48 @@
...
+import { guidFor } from '@ember/object/internals';
+import { tracked } from '@glimmer/tracking';
+import { triplesForPath } from '@lblod/submission-form-helpers';
+import { SKOS } from '@lblod/submission-form-helpers';
+import rdflib from 'browser-rdflib';
+
+export default class RdfInputFieldsConceptSchemeRadioButtonsShowComponent extends InputFieldComponent {
+  inputId = 'conceptSchemeRadioButtons-' + guidFor(this);
+
+  @tracked selected = null
```
HugaertsDries, line 11: Here you kept them :thinking:

addon/components/rdf-input-fields/concept-scheme-radio-buttons/show.js
```diff
@@ -0,0 +1,48 @@
...
+
+  loadProvidedValue() {
+    if (this.isValid) {
+      // Assumes valid input
+      // This means even though we can have multiple values for one path (e.g. rdf:type)
+      // this selector will only accept one value, and we take the first value from the matches.
+      // The validation makes sure the matching value is the sole one.
+      const matches = triplesForPath(this.storeOptions, true).values;
+      this.selected = this.options.find(opt => matches.find(m => m.equals(opt.subject)));
+      console.log(this.options)
```
HugaertsDries, line 43: :warning: best to remove this before releasing

addon/components/rdf-input-fields/concept-scheme-radio-buttons/edit.js
aatauil: Codes comes from concept-scheme-selector as idea is pretty similar to radio-buttons: "multiple options to select from and only 1 can get selected at a time." So assumed that radio-buttons would have same flow , my assumption was right  apparently :D

addon/components/rdf-input-fields/concept-scheme-radio-buttons/show.hbs
```diff
@@ -0,0 +1,16 @@
+<AuLabel class="{{if this.errors "au-c-label--error"}}" for={{this.inputId}}>
+  {{@field.label}}
+  {{#if this.isRequired}}
+    <AuPill>Verplicht</AuPill>
+  {{/if}}
+</AuLabel>
+
+{{#each this.options as |option|}}
+  <label class="radio">
+    <input type="radio" class="radio__toggle" disabled={{true}} checked={{if (eq option.label option.selected.label) "true" "false"}} name={{@field.uri}} value={{option}}>
```
HugaertsDries, line 10: Does this check work? I never seem to find where you put the selected on an `option`, within the `options`? Also, wouldn't it be better to mach on the `subject` instead off the `label`?

aatauil: Yes this actually fixes the  issue I had. Will push in a minute

HugaertsDries (approved):

PR #13, opened 2021-03-30
Feature/multi-select-checkboxes

HugaertsDries:
Personally not the biggest fan off the name, better suggestions are welcome.


claire-lovisa (approved):
Tested and approved :)

PR #17, opened 2021-05-03
Improvement/accesibility

aatauil:
- Mainly improve accessibility for the tables by adding scope to it and fixing small other things.
- note: the objective table is a "complex table" so it needed another method for making it accessible then the other tables.


claire-lovisa:
As you know I had troubles testing, but code-wise it looks good at least :sweat_smile:

PR #19, opened 2021-05-11
Remove unneeded / nonexistent exports

Windvis:
These components were probably renamed but the app files weren't removed. The show.js module doesn't exist, which makes the Embroider build fail.


aatauil:
Indeed, those come from the older radio-buttons we had and the exports must have been left in. Good catch !

PR #20, opened 2021-05-11
Use the {{ensure-safe-component}} helper

Windvis:
This makes the helpers statically analyzable which means stricter Embroider flags can be enabled.

More information: https://github.com/embroider-build/embroider/blob/master/REPLACING-COMPONENT-HELPER.md

### DIFF
--- a/addon/components/property-group.hbs
+++ b/addon/components/property-group.hbs
@@ -22,7 +22,7 @@
     {{#each this.children as |child index|}}
       {{#unless (is-property-group child.displayType)}}
         <div class="au-u-margin-bottom-small">
-          {{component (component-for-display-type child.displayType show=@show)
+          {{component (ensure-safe-component (component-for-display-type child.displayType show=@show))
                       field=child
                       form=@form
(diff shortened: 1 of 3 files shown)


aatauil:
Summary log: 

-  Shows deprecations because we pass string instead of imported component 
   -> can be easily fixed by importing all components for the type mapping 
 
 - Empty string return logic in component-for-display-type helper causes component helper to return error 
  -> find out why empty string gets returned
  -> depending on above answer add missing mappings or find workaround

Windvis:
I'll put this one on hold for now. I didn't know about the deprecations. I'll update that later and fix the other issue in a separate PR.

Windvis:
I'm going to close this. If we drop Ember 3.24 support we can just invoke the components directly without the need to pass a dynamic value to the component helper which  makes this PR unneeded.

Windvis:
I'm going to close this. If we drop Ember 3.24 support we can just invoke the components directly without the need to pass a dynamic value to the component helper which  makes this PR unneeded.

PR #22, opened 2021-05-14
Don't render form fields without a displayType

Windvis:
Some forms contain fields which are used for validation but don't need to be rendered.

I also added asserts that shows an error when a mapping isn't found. It might be easier to track missing mappings that way.

And I fixed some linting errors that I encountered.

Fixes the errors from #20

### DIFF
--- a/addon/components/property-group.hbs
+++ b/addon/components/property-group.hbs
@@ -20,39 +20,41 @@
     {{/if}}
     {{!-- NOTE: Children can be property-groups as well as fields --}}
     {{#each this.children as |child index|}}
-      {{#unless (is-property-group child.displayType)}}
-        <div class="au-u-margin-bottom-small">
-          {{component (component-for-display-type child.displayType show=@show)
-                      field=child
-                      form=@form
-                      formStore=@formStore
-                      graphs=@graphs
-                      sourceNode=@sourceNode
-                      forceShowErrors=@forceShowErrors
-                      cacheConditionals=@cacheConditionals
-                      show=@show
-          }}
-          <AuHelpText class="au-u-margin-none">{{child.help}}</AuHelpText>
-        </div>
-      {{else}}
-        <PropertyGroup
-                @level={{add this.level 1}}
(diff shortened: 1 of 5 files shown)


addon/components/property-group.hbs
```diff
@@ -46,13 +32,27 @@
                 @cacheConditionals={{@cacheConditionals}}
                 @show={{@show}}/>
         {{#if @last}}
-          {{#unless (eq index (sub this.children.length 1))}}
+          {{#if (not (eq index (sub this.children.length 1)))}}
```
aatauil, line 35: Just a small detail but maybe we should use the not-eq helper instead here?
`{{#if (not-eq index (sub this.children.length 1))}}`

aatauil (commented):
I tested the changes and noticed an issue with one field. When you navigate to Toezicht and click on "Besluitenlijst" then the browser crashes in my case. (if it renders correctly the first time choose something else in the list then click again on "besluitenlijst"). I am not 100% sure these changes are the cause of it but on master I havent been able to reproduce the same bug yet so could you let me know how it goes for

Windvis:
For some reason the observer goes into an infinite loop and I don't understand why. I rewrote the logic to use the ember array version, but it still happens there. I think it's just my lack of background context about how this all works together.

Loop goes like this:

- update fields adds missing fields
- fields without a displayType get removed
- loop
- update fields adds the previously removed fields again
- fields get removed
- loop 
- ...

I'll go for the template version for now.


I'm gonna try the template version instead.

aatauil:
@Windvis is the loop similar to what is described in this issue? 
https://github.com/lblod/ember-submission-form-fields/issues/29

Windvis:
@aatauil Good question, it's been a while since I did the debugging :smile:. I also force pushed so I can't remember what the previous implementation looked like :smile:.

Chances are high it was indeed related to that issue though.

Windvis:
Closing since I'll tackle integrate it as part of #41

PR #23, opened 2021-05-18
Feature/climate accordion

aatauil:


addon/components/custom-subsidy-form-fields/climate-subsidy-costs-table/edit.hbs
```diff
@@ -54,8 +54,16 @@
         as |row|>
 
         <row.actionDescription>
-          <label class="au-c-label ">Algemene beleidsdoelstellingen:</label>
-          Opmaak, uitvoering en opvolging Burgemeestersconvenant actieplan 2030
+          <AuAccordion
+            @accordionIconOpen="nav-up"
```
Windvis, line 58: I find it weird that this needs to be repeated everywhere. `nav-up` and `nav-down` seem like sane default values for these arguments. 

Anyway, I would wrap this in a project specific wrapper that sets these defaults up. And maybe alias some args as well?
- `accordionButtonLabel` -> `label`
- `accordionSubTitle` -> `subtitle`

aatauil: Jup that would indeed be logical but for some reason the default value's are empty for the icons. The args should actually never be called that, no other component that I  know of prepends its name for all arguments, I will open an issue for that as It doesnt make sense in my eyes

aatauil: Can you elaborate on what you mean with project specific wrapper ? nested components ?

Windvis: Yea just a template only component that wraps AuAccordion and passes through the defaults.

Something like this:
```hbs
{{! accordion.hbs}}
<AuAccordion
  @accordionIconOpen="nav-up"
  @accordionIconClosed="nav-down"
  @accordionButtonLabel={{@label}}
  @accordionSubTitle={{@subtitle}}
>
  {{yield}}
</AuAccordion>
```

Which can then be used like
```hbs
<Accordion
  @label="Some label"
  @subtitle="Some subtitle"
> Content </Accordion>
```

Windvis: Good stuff!

FYI in case you don't know. If you create a template only component in an addon, and the consuming app still hasn't enabled the "template-only-glimmer-components" optional feature, Ember will generate a classic component as the backing class. So it's generally a good practice to [export a templateOnly component explicitly](https://github.com/emberjs/ember.js/issues/19325).

For this addon it doesn't really matter, since the consuming apps are all using Octane already (I assume), but it might come in handy if you make more generic addons.

aatauil: Aha didnt know that , yes the consuming app is indeed already fully octane but I will remember that , thanks for the heads-up!

PR #25, opened 2021-07-13
add logic to distinguish if form is 'aanvraagStep'

aatauil:
See DL-2814 for more info

### DIFF
--- a/addon/components/custom-subsidy-form-fields/estimated-cost-table/edit.js
+++ b/addon/components/custom-subsidy-form-fields/estimated-cost-table/edit.js
@@ -116,21 +116,12 @@ export default class CustomSubsidyFormFieldsEstimatedCostTableEditComponent exte
   }
 
   get isAanvraagStep() {
-    const results = this.storeOptions.store.match(
-      undefined,
-      optionsPredicate,
-      undefined,
-      undefined);
-
-    const optionsList = results.map(item => {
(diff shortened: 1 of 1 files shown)


addon/components/custom-subsidy-form-fields/estimated-cost-table/edit.js
```diff
@@ -97,6 +115,24 @@ export default class CustomSubsidyFormFieldsEstimatedCostTableEditComponent exte
                                           this.storeOptions.sourceGraph).length > 0;
   }
 
+  get isAanvraagStep() {
+    const results = this.storeOptions.store.match(
+      undefined,
+      optionsPredicate,
+      undefined,
+      undefined);
```
claire-lovisa, line 123: We can use `this.args.field.options` instead (as discovered while chatting ;))

claire-lovisa (changes requested):
A small change with `this.args.field.options` that'll simplify the code, other than that seems perfect !

addon/components/custom-subsidy-form-fields/estimated-cost-table/edit.js
aatauil: Changed it to use this.args, also testing if this.args.field && this.args.field.options exist just to be sure

claire-lovisa (approved):
All good thanks :D

PR #28, opened 2021-09-10
fix issue of being able to submit when costTable cost field is empty

aatauil:


addon/components/custom-subsidy-form-fields/estimated-cost-table/edit.js
```diff
@@ -372,7 +382,13 @@ export default class CustomSubsidyFormFieldsEstimatedCostTableEditComponent exte
         this.updateTripleObject(this.estimatedCostTableSubject, validEstimatedCostTable, true);
       }
 
-      this.updateTripleObject(entry.estimatedCostEntrySubject, entry['cost'].predicate, entry['cost'].value);
+
+      if(isNaN(parseInt(entry.cost.value))) {
+        this.updateTripleObject(entry.estimatedCostEntrySubject, entry['cost'].predicate, "Field is empty");
```
aatauil, line 387: If value of cell NaN then make the object of that triple a string with a value. When you input something else like "", undefined, null, false and you later on look for that triple see line 345, you wont get that triple in the results and theirfor cant validate if a field was left empty

addon/components/custom-subsidy-form-fields/estimated-cost-table/edit.js
```diff
@@ -363,7 +373,7 @@ export default class CustomSubsidyFormFieldsEstimatedCostTableEditComponent exte
     updateCost(entry){
       entry.cost.errors = [];
 
-      if (!this.isPositiveInteger(Number(entry.cost.value))) {
+      if (!this.isPositiveInteger(entry.cost.value)) {
```
aatauil, line 376: Remove Number() as it will otherwise not show the error message when a field is left empty

addon/components/custom-subsidy-form-fields/estimated-cost-table/edit.js
```diff
@@ -348,13 +348,23 @@ export default class CustomSubsidyFormFieldsEstimatedCostTableEditComponent exte
...
+    const positiveCosts = entries.filter(entry => parseInt(entry.object.value) > 0 );
+    if(!positiveCosts.length){
       this.errors.pushObject({
         message: 'Mintens één kosten veld moet een waarde groter dan 0 bevatten.'
       });
       this.updateTripleObject(this.estimatedCostTableSubject, validEstimatedCostTable, null);
-    } else {
+    }
+
+    if(positiveCosts.length && !inValidCosts.length) {
```
aatauil, line 367: if, else if, else doesnt work here as we want both positiveCost & inValidCost to be correct

addon/components/custom-subsidy-form-fields/estimated-cost-table/edit.js
```diff
@@ -348,13 +348,23 @@ export default class CustomSubsidyFormFieldsEstimatedCostTableEditComponent exte
       this.storeOptions.sourceGraph
     );
 
-    const validCosts = entries.filter(entry => parseInt(entry.object.value) > 0 );
-    if(!validCosts.length){
+    const inValidCosts = entries.filter(entry => isNaN(parseInt(entry.object.value)));
+    if(inValidCosts.length) {
```
claire-lovisa, line 352: very small detail, but shouldn't it be `invalidCosts` instead of `inValidCosts` ? Made me confused for a sec :see_no_evil:

addon/components/custom-subsidy-form-fields/estimated-cost-table/edit.js
```diff
@@ -348,13 +348,23 @@ export default class CustomSubsidyFormFieldsEstimatedCostTableEditComponent exte
...
+      this.errors.pushObject({
+        message: 'Eén van de velden werd niet correct ingevuld.'
+      });
+      this.updateTripleObject(this.estimatedCostTableSubject, validEstimatedCostTable, null);
```
claire-lovisa, line 356: Why is it set to `null` and not `false` here ?

addon/components/custom-subsidy-form-fields/estimated-cost-table/edit.js
claire-lovisa: Couldn't we set it to 0 directly if not a number ? Like ... force-set the value instead of letting it be a NaN, that would ease a bit process because we wouldn't have to check for inValidCosts ?

addon/components/custom-subsidy-form-fields/estimated-cost-table/edit.js
aatauil: Oh my camelCase mind bugged out there, yeah indeed invalid is one word :p

addon/components/custom-subsidy-form-fields/estimated-cost-table/edit.js
aatauil: Very good question, I actually considered changing it all to false but then realized that other components (see objectivesTable) also uses null so kept it for the sake of consistentie. Better to make a separate PR changing this everywhere because I am not sure if this will actually have side effects

addon/components/custom-subsidy-form-fields/estimated-cost-table/edit.js
aatauil: Yeah I actually mentioned it to Felix, this is debatable since users wont know that an empty field is equal to 0 so I thought that this could cause confusion (especially if you are talking about money) so better to let the user put a 0 so they know what to expect

claire-lovisa (approved):

PR #30, opened 2021-10-06
Fix/empty tables | climate & objectives table

aatauil:
Instead of using next() creating inconsistencies and only calculating values after render we now use scheduleOnce and move the function call to be executed in the current runloop in the very first queue (actions queue). 

This also removes all warnings in console about re-computing


HugaertsDries (approved):
Nice find

PR #34, opened 2022-02-22
Fix linting errors

Windvis:
Most of these are autofixes I recommend checking the js and hbs commits when reviewing.


tests/dummy/app/routes/application.js
```diff
@@ -0,0 +1,53 @@
+import Route from '@ember/routing/route';
```
Windvis, line 1: this shouldn't have been commited. woops.

PR #36, opened 2022-03-01
Modernize the built-in components

Windvis:
Some components were still using older appuniversum patterns (things like custom error states, manual icon placement, required labels, etc). These are replaced by the modern equivalents provided by ember-appuniversum v1.0.

- The built-in error, and required arguments from the label component should be used instead of the custom version
- show components should not display the errors and required labels
- use semantically correct html if appuniversum supports it already


addon/components/file-card.hbs
```diff
@@ -0,0 +1,38 @@
+<div
```
Windvis, line 1: This should ideally be moved to appuniversum. This isn't strictly related to the form fields, so it feels out of place in this addon.

addon/components/file-card.hbs
```diff
@@ -0,0 +1,38 @@
...
+    >
+      <AuIcon @icon="cross" @size="large" />
+      <span class="au-u-hidden-visually">Verwijderen</span>
+    </button>
+  {{/if}}
+
+  <div class="au-o-box au-o-box--small au-c-card">
+    <AuContent @skin="small">
+      <AuHeading @level="4" @skin="6">
+        {{@file.filename}}
```
Windvis, line 21: Maybe the file shouldn't be passed in directly so we aren't stuck with a specific file model? (or being forced to build the file object before passing it in).

Windvis:
While testing out this branch in Loket it seems console errors are thrown when more than 1 file-upload component is rendered:
`Assertion Failed: Cannot add multiple listeners for the same element #ember479, #ember472`

This isn't happening in the dev environment so I'm guessing it's related to the ember-file-upload v4 update.

addon/components/file-card.hbs
Windvis: https://github.com/appuniversum/ember-appuniversum/issues/213

PR #39, opened 2022-03-02
Refactor the date components

Windvis:
This modernizes the date components:
- remove usage of webuniversum components
- remove usage of moment

### DIFF
--- a/addon/utils/date.js
+++ b/addon/utils/date.js
@@ -1,0 +1,10 @@
+export function formatDate(date) {
+  if (!(date instanceof Date) || isNaN(date)) {
+    return '';
+  }
+
+  let day = `${date.getDate()}`.padStart(2, '0');
+  let month = `${date.getMonth() + 1}`.padStart(2, '0');
+
+  return `${day}.${month}.${date.getFullYear()}`;
+}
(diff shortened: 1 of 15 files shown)


Windvis:
Works as expected in loket. It only has some styling conflicts because of the generic table styles in ember-vo-datatable but that will be removed anyways.

addon/utils/date.js
```diff
@@ -0,0 +1,10 @@
+export function formatDate(date) {
+  if (!(date instanceof Date) || isNaN(date)) {
+    return '';
+  }
+
+  let day = `${date.getDate()}`.padStart(2, '0');
```
aatauil, line 6: Great find on the padStart function! I didnt know that existed, would have been handy on other components like the timepicker that I created in appuniversum :p

aatauil:
Other than the style conflict in loket that you mentioned above, I did not find any other issues and everything seems to work as expected :)

Windvis:
That can be fixed by commenting out `@import 'ember-vo-webuniversum-data-table'; // Deprecated` in the app.scss file, but I didn't want to spend time to find out if it affected other things so I didn't remove that file in this PR. Once all webuniversum components and styles are removed we can do a sweep over the app and find out where things look broken.

PR #44, opened 2022-03-09
Remove webuniversum dependency

Windvis:
This removes the dependency on ember-vo-webuniversum since we don't use it anymore. I temporarily copied the styles so that the tables look as they should until we find the time to refactor them to newer patterns.

This includes the commits from #36 so that one needs to be merged first.


Windvis:
I've included the commits in the other PR.

PR #48, opened 2022-03-22
Optionally add new actions to the climate subsidy table

Windvis:
image

Based on #53 so that I could deduplicate the component already (and not have to repeat the changes in the show component), so that one needs to be merged first.


addon/components/custom-subsidy-form-fields/climate-subsidy-costs-table/edit.js
```diff
@@ -206,6 +233,30 @@ export default class CustomSubsidyFormFieldsClimateSubsidyCostsTableEditComponen
     }
   }
 
+  @action
+  addCustomData() {
+    let uuid = uuidv4();
+    const tableEntryUri = new rdflib.NamedNode(
+      `http://data.lblod.info/id/climate-table/row-entry/${uuid}`
+    );
```
Windvis, line 241: TL;DR do we need to store the uuid in the store?

This is were we generate the uri for new custom table entries. The idea is that the custom-row component can then use this uri to check if data already exists and if not, add default values. The quirky thing is that the other table rows all add a uuid triple, but to do that here we would also need to pass the uuid to the component.

This complicates things if we want to fetch the existing custom rows, since we would also need to fetch the uuid values.

addon/components/custom-subsidy-form-fields/climate-subsidy-costs-table/edit.js
```diff
@@ -206,6 +233,30 @@ export default class CustomSubsidyFormFieldsClimateSubsidyCostsTableEditComponen
...
+
+    this.customDataEntries = [
+      ...this.customDataEntries,
+      {
+        uri: tableEntryUri,
+        uuid,
+        temp: `http://data.lblod.info/id/subsidies/rules/custom/${uuid}`,
```
Windvis, line 248: This was temporarily passed here since I based the custom-data component on the other row components and they expect this to be passed.

Do we need a business rule for the custom entries? If so, should we just make it static?

addon/components/custom-subsidy-form-fields/climate-subsidy-costs-table/table-row-custom-data.js
```diff
@@ -0,0 +1,407 @@
...
+    return new rdflib.NamedNode(this.args.businessRuleUriStr);
+  }
+
+  get climateTableSubject() {
+    return this.args.climateTableSubject;
+  }
+
+  // TODO: this logic probably needs to be adjusted
+  get hasValues() {
+    const values = this.storeOptions.store.match(
```
Windvis, line 78: The plan is to pass the tableEntry uri from the parent component and use that uri to check for values. Is that ok?

addon/components/custom-subsidy-form-fields/climate-subsidy-costs-table/table-row-custom-data.js
```diff
@@ -0,0 +1,407 @@
...
+        predicate: RDF('type'),
+        object: ClimateEntryType,
+        graph: this.storeOptions.sourceGraph,
+      },
+      {
+        subject: tableEntryUri,
+        predicate: MU('uuid'),
+        object: uuid,
+        graph: this.storeOptions.sourceGraph,
+      },
```
Windvis, line 167: As mentioned before, this is the reason why we would need to pass in the uuid from the parent component, but I'm not sure if it's really needed since we won't use mu-cl-resources to fetch this?

addon/components/custom-subsidy-form-fields/climate-subsidy-costs-table/table-row-custom-data.js
```diff
@@ -0,0 +1,407 @@
...
+      {
+        subject: tableEntryUri,
+        predicate: climateEntryCustomData,
```
Windvis, line 170: When loading a table we need a way to find the custom table entries so we can show the previously saved rows again. I opted for a custom predicate with a boolean values, but there probably are better ways of doing it? A custom type maybe? `CustomClimateEntry`?

Windvis:
@cecemel is the current workflow what you had in mind? The jira issue is vague in what's expected.

The idea is to show that button below the table. Users click it if they want to add extra actions, and they can remove the rows again until the form is submitted. (So the show version will only display the persisted rows).

Windvis:
Update after the discussion yesterday:
- We should stay consistent with the other table rows and generate a business uri for each new table row
- We should add an order so the rows are rendered in the same order they were added
- We need to add some sort of predicate to know that it's a custom row (this also needs to be configured in the backend somewhere)

addon/components/custom-subsidy-form-fields/climate-subsidy-costs-table/table-row-custom-action.js
```diff
@@ -0,0 +1,416 @@
...
+const amountPerActionPredicate = new rdflib.NamedNode(
+  `${climateBaseUri}amountPerAction`
+);
+const restitutionPredicate = new rdflib.NamedNode(
+  `${climateBaseUri}restitution`
+);
+const hasInvalidRowPredicate = new rdflib.NamedNode(
+  `${climateTableBaseUri}/hasInvalidClimateTableEntry`
+);
+const customDescriptionPredicate = new rdflib.NamedNode(
```
Windvis, line 42: This is the predicate that is used to store the custom description text, but there might be a better way of doing this? a preflabel for the businessRule maybe?

addon/components/custom-subsidy-form-fields/climate-subsidy-costs-table/table-row-custom-action.js
```diff
@@ -0,0 +1,416 @@
...
+const climateBaseUri = 'http://data.lblod.info/vocabularies/subsidie/climate/';
+const climateTableBaseUri = 'http://data.lblod.info/climate-tables';
+
+const tableEntryBaseUri = 'http://data.lblod.info/id/climate-table/row-entry';
+const ClimateEntryType = new rdflib.NamedNode(`${climateBaseUri}ClimateEntry`);
+const climateEntryPredicate = new rdflib.NamedNode(
+  `${climateBaseUri}climateEntry`
+);
+const climateEntryCustomAction = new rdflib.NamedNode(
+  `${climateBaseUri}customAction`
```
Windvis, line 28: This is added to the custom rows so that we can retrieve the data in the parent component on load. Let me know if there are better ways of doing this or if I should use a different name.

addon/components/custom-subsidy-form-fields/climate-subsidy-costs-table/table-row-custom-action.js
```diff
@@ -0,0 +1,416 @@
...
+        message: 'Aantal items moeten een geheel getal vormen.',
+      });
+
+      return false;
+    }
+
+    //OpstartTraject lokale energiegemeentschap
+    else if (
+      toRealiseUnits > 1 &&
+      'http://data.lblod.info/id/subsidies/rules/38d6d2bd-e42b-4d7e-8fea-9a371d9cf22f' ==
```
Windvis, line 380: Woops, this can be removed.

PR #51, opened 2022-03-24
Fix the heading levels

Windvis:
We were passing a number into the `@level` argument of the `AuHeading` component. This component uses the `eq` helper to compare the values which is a strict equal so non of the values matched and it always defaulted to a h1. `{{eq 2 "2"}}` will return `false`.

This change should result in the proper title being used.

### DIFF
--- a/addon/components/property-group.js
+++ b/addon/components/property-group.js
@@ -34,6 +34,18 @@ export default class SubmissionFormPropertyGroupComponent extends Component {
     return this.args.level || 1;
   }
 
+  get titleLevel() {
+    return `${this.level}`;
+  }
+
+  get titleSkin() {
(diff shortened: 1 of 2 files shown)


addon/components/property-group.js
```diff
@@ -34,6 +34,18 @@ export default class SubmissionFormPropertyGroupComponent extends Component {
     return this.args.level || 1;
   }
 
+  get titleLevel() {
+    return `${this.level}`;
+  }
+
+  get titleSkin() {
+    return `${this.level + 1}`;
```
Windvis, line 42: I've used a getter here as well since that brings us 1 step closer to removing ember-math-helpers addon. (only `sub` remains I think).

PR #52, opened 2022-03-25
Parse field options in the field model

Windvis:
I noticed a lot of components were using JSON.parse to read the field options. We can simplify all this code by doing the work in the field model instead.


Windvis:
This will cause conflicts with the deduplicate branch so it's probably better to rebase after that one is merged.

PR #53, opened 2022-03-28
Display type mapping improvements

### LINKED ISSUE #41: Allow consuming applications to register custom form fields
At the moment the display types are mapped to component names by a mapping that is stored in the helpers. It would be nice if this mapping wasn't hardcoded so applications could add new mappings without the components having to be part of this addon. Components that are only used in a single app don't need to be shipped to all applications that consume this addon.

The helpers itself could be refactored as well. At the moment the mapping is duplicated.

Something like this could work as a public API:
```diff
// app/app.js

import Application from '@ember/application';
import Resolver from 'ember-resolver';
import loadInitializers from 'ember-load-initializers';
import config from './config/environment';
import './models/custom-inflector-rules';
+ import { registerCustomFields } from '@lblod/ember-submission-form-fields';
+ import SomeEditComponent from 'app-name/components/some-edit-component';
+ import SomeShowComponent from 'app-name/components/some-show-component';

export default class App extends Application {
  modulePrefix = config.modulePrefix;
  podModulePrefix = config.podModulePrefix;
  Resolver = Resolver;
}

loadInitializers(App, config.modulePrefix);
+ registerCustomFields([{
+   displayType: 'http://lblod.data.gift/display-types/some-display-type',
+   editComponent: SomeEditComponent,
+   showComponent: SomeShowComponent
+ }]);
```

Alternatively a service could be used to store the mapping.

The components themself heavily use inheritance to share functionality. While I don't think that's the best solution, refactoring that away is a huge undertaking. Custom components will also need some of that shared logic, so it probably makes sense to create a base class that can be considered public API as well.

`import FormField from '@lblod/ember-submission-form-fields';`

### LINKED ISSUE #62: Embroider support
The addon uses the component helper to render the form field components. The display types are mapped to a component name and returned by a helper. This is something that Embroider can't analyze so it will throw an error.

We should change the mapping so that it returns the component classes instead of the component name. That way we can drop the component helper altogether. Rendering component classes directly is only supported in Ember 3.25+ but we can support older version by wrapping the class in a `ensureSafeComponent` call from `@embroider/utils`.

### LINKED ISSUE #63: simplify the display type mapping
At the moment 3 helpers are used to map a display type to a component name. 
- component-for-display-type-edit
- component-for-display-type-show
- component-for-display-type

Each helper also has an internal mapping so adding new display types means we need to edit multiple configs. This has the risk of things getting out of sync.

We should just drop the -edit and -show helper and use the `component-for-display-type` helper everywhere. This means we only need a single map for the edit and show components.

Windvis:
This makes the addon Embroider compatible and also stops rendering fields without a display type which created unnecessary white space in the Toezicht module.

Before:
image

After:
image

Closes #63 
Closes #62


addon/utils/display-type.js
```diff
@@ -0,0 +1,155 @@
+import { assert } from '@ember/debug';
+
+// Basic fields
+import BestuursorgaanSelectorEditComponent from '@lblod/ember-submission-form-fields/components/custom-submission-form-fields/bestuursorgaan-selector/edit';
```
Windvis, line 4: Huge list of imports, but once we deduplicate the components it will be halved and it will be further reduced once we implement the component registration functionality and the custom tables can be extracted.

addon/components/property-group.hbs
```diff
@@ -20,40 +20,45 @@
...
-          <AuHelpText>{{{child.help}}}</AuHelpText>
-        </div>
-      {{else}}
-        <PropertyGroup
-                @level={{this.nextLevel}}
+    {{#each this.children as |field index|}}
+      {{#if field.displayType}}
+        {{#if (not (is-property-group field.displayType))}}
+          <div class="au-u-margin-bottom-small">
+            {{#let (component-for-display-type field.displayType show=@show) as |FormField|}}
```
Windvis, line 27: This won't work in < Ember 3.25 so in theory it's a breaking change. We can wrap it in an ensure-safe-component to make it work in older versions so we can release it as a 1.X release. In 2.0 we can then drop 3.25 support and remove ensure-safe-components again.

addon/utils/display-type.js
```diff
@@ -0,0 +1,166 @@
...
+import CheckboxEditComponent from '@lblod/ember-submission-form-fields/components/rdf-input-fields/checkbox/edit';
+import CheckboxShowComponent from '@lblod/ember-submission-form-fields/components/rdf-input-fields/checkbox/show';
+import ConceptSchemeRadioButtonsEditComponent from '@lblod/ember-submission-form-fields/components/rdf-input-fields/concept-scheme-radio-buttons/edit';
+import ConceptSchemeRadioButtonsShowComponent from '@lblod/ember-submission-form-fields/components/rdf-input-fields/concept-scheme-radio-buttons/show';
+import ConceptSchemeSelectorEditComponent from '@lblod/ember-submission-form-fields/components/rdf-input-fields/concept-scheme-selector/edit';
+import ConceptSchemeSelectorShowComponent from '@lblod/ember-submission-form-fields/components/rdf-input-fields/concept-scheme-selector/show';
+import ConceptSchemeMultiSelectorEditComponent from '@lblod/ember-submission-form-fields/components/rdf-input-fields/concept-scheme-radio-buttons/edit';
+import ConceptSchemeMultiSelectorShowComponent from '@lblod/ember-submission-form-fields/components/rdf-input-fields/concept-scheme-radio-buttons/show';
```
Windvis, line 15: These imports are wrong.

addon/components/submission-form.hbs
```diff
@@ -7,28 +7,27 @@
...
-          {{! template-lint-disable no-triple-curlies}}
-          <AuHelpText>{{{field.help}}}</AuHelpText>
-        {{/if}}
-      </li>
+            <FormField
+              @field={{field}}
+              @formStore={{@formStore}}
+              @graphs={{@graphs}}
+              @sourceNode={{@sourceNode}}
+              @forceShowErrors={{@forceShowErrors}}
```
Windvis, line 22: `@show` needs to be passed to the component.

addon/components/submission-form.hbs
```diff
@@ -7,28 +7,27 @@
...
-              (component-for-display-type-edit field.displayType)
-              field=field
-              formStore=@formStore
-              graphs=@graphs
-              sourceNode=@sourceNode
-              forceShowErrors=@forceShowErrors
+      {{#if field.displayType}}
+        <li class="au-o-grid__item au-u-1-1">
+          {{#let (ensure-safe-component
+            (component-for-display-type field.displayType @show))
```
Windvis, line 14: Needs to be passed as a named argument
```suggestion
            (component-for-display-type field.displayType show=@show))
```

addon/helpers/component-for-display-type-edit.js
```diff
@@ -1,36 +0,0 @@
-import { helper } from '@ember/component/helper';
```
Windvis, line 1: The app file needs to be removed as well.

addon/components/property-group.hbs
```diff
@@ -20,40 +20,48 @@
...
-          }}
-          {{! template-lint-disable no-triple-curlies}}
-          <AuHelpText>{{{child.help}}}</AuHelpText>
-        </div>
-      {{else}}
-        <PropertyGroup
-                @level={{this.nextLevel}}
+    {{#each this.children as |field index|}}
+      {{#if field.displayType}}
+        {{#if (not (is-property-group field.displayType))}}
```
aatauil, line 25: Wouldnt it be nicer to remove `(not` and just switch the conditions around?

```
  {{#if (is-property-group field.displayType)}}
     <PropertyGroup
        @level={{this.nextLevel}}
        @form={{@form}}
        @group={{field}}
        @formStore={{@formStore}}
        @graphs={{@graphs}}
        @sourceNode={{@sourceNode}}
        @forceShowErrors={{@forceShowErrors}}
        @cacheConditionals={{@cacheConditionals}}
        @show={{@show}}
      />
  
      {{#if @last}}
        {{#if (not-eq index (sub this.children.length 1))}}
          <AuHr @size="large"/>
        {{/if}}
      {{else}}
        <AuHr @size="large"/>
      {{/if }}

  {{else}}
    <div class="au-u-margin-bottom-small">
      {{#let (ensure-safe-component
        (component-for-display-type field.displayType show=@show))
        as |FormField|
      }}
        <FormField
          @field={{field}}
          @form={{@form}}
          @formStore={{@formStore}}
          @graphs={{@graphs}}
          @sourceNode={{@sourceNode}}
          @forceShowErrors={{@forceShowErrors}}
          @cacheConditionals={{@cacheConditionals}}
          @show={{@show}}
        />
      {{/let}}

      {{! template-lint-disable no-triple-curlies}}
      <AuHelpText>{{{field.help}}}</AuHelpText>
    </div>
  {{/if}}
        
```

Windvis: Yes, I just adjusted what was already there. I'll swap them around :+1:

aatauil (commented):
Looks great! kinda afraid that display-type list might become huge in the far future other then that everything still works as should :)

addon/components/property-group.hbs
Windvis: Should be good now.

aatauil: Thanks! :smile:

Windvis:
I'm merging this since it blocks the other PRs. I did another test run in both the toezicht and subsidy modules in loket and everything seems to work as expected.

PR #54, opened 2022-03-29
Deduplicate field components

### LINKED ISSUE #50: Deduplicate the templates
It seems that most `edit`/`show` combinations use mostly the same template with some minor deviations. Maybe it makes sense to default to only using the edit component and pass in a `@readOnly` property or something? The components that are way different can still use a dedicated show component.

It would reduce the duplication of code and results in smaller bundle sizes since there are less templates to compile.

Windvis:
~~This is based on #53 so that should be merged first.~~ Rebased

Closes #50


aatauil (approved):
LGTM, went through all subsidy forms and some toezicht forms and everything works fine. Thanks for the great work!

PR #56, opened 2022-03-30
Delete the "burgemeester" row of the climate subsidy table

Windvis:
This row was already commented out in the `edit` component but still visible in the `show` component.

I also deleted the table-row-burgemeester component since it's no longer used.


Windvis:
Draft until we know for sure that deleting is the way to go.

Windvis:
It seems deleting isn't the right call and instead we should enable the row again in the edit version.

PR #71, opened 2022-04-19
Ensure checkboxes show the correct checked state when data is loaded

### LINKED ISSUE #67: [Checkbox] checked + disabled
Low prio reminder: The checked state should be passed as an argument (checked -> @checked). @Windvis unless there is reason its done this way? Currently, because of this, when a checkbox is both checked + disabled it will only show the disabled state. This can cause some confusion/panic for the user

https://github.com/lblod/ember-submission-form-fields/blob/master/addon/components/rdf-input-fields/checkbox.hbs#L12
https://github.com/lblod/ember-submission-form-fields/blob/master/addon/components/rdf-input-fields/concept-scheme-multi-select-checkboxes.hbs#L26

Windvis:
`AuControlCheckbox` uses the `Input` component internally which requires `checked` to be passed as an argument instead of an attribute.

Fixes #67

### DIFF
--- a/addon/components/rdf-input-fields/checkbox.hbs
+++ b/addon/components/rdf-input-fields/checkbox.hbs
@@ -10,7 +10,7 @@
   @label={{@field.label}}
   @disabled={{@show}}
   @onChange={{this.updateValue}}
-  checked={{this.value}}
+  @checked={{this.value}}
 />
 {{#each this.errors as |error|}}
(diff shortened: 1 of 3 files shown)


Windvis:
I verified that the problem is resolved by linking the addon in loket. Checkboxes are checked again when the page is reloaded :+1:.

PR #74, opened 2022-04-27
Use MSW to mock the case-number request

Windvis:
MSW is framework agnostic and better documented than Ember's built-in mocking functionality.


tests/dummy/app/mocks/handlers.js
```diff
@@ -0,0 +1,9 @@
+import { rest } from 'msw';
+
+export const handlers = [
+  rest.post('/case-number-generator/generate', (req, res, ctx) => {
+    const prefix = req.url.searchParams.get('prefix');
+
+    return res(ctx.status(200), ctx.json([`${prefix}1234`]));
```
Windvis, line 7: This returns a hardcoded case-number, but that's ok for the test app.

PR #77, opened 2022-05-20
[WIP] Create a collapsibe property  group variant

Windvis:
Some quick experimentation to see if it would be possible to create a "collapsible" property group variant.

It needs more work.


Windvis:
Superseded by #148

PR #78, opened 2022-05-26
Test/add editor as form field

lagartoverde:


lagartoverde:
I think it's good enough like this, please review and check with direction I should go with this, if you want to run in loket you would need the branch of the same name in both the frontend and the backend. I didn't create a PR there because I think they are not needed yet

lagartoverde:
Are test usually passing? because I get a problem in one and I don't know what's happening there

addon/components/rdfa-editor-container.hbs
```diff
@@ -0,0 +1,45 @@
+{{#if @busy}}
```
Windvis, line 1: This doesn't seem to be used? If it's not needed I would delete it since it seems to be some sort of implementation detail of the editor that was copied over?

addon/components/rdf-input-fields/rich-text.hbs
```diff
@@ -0,0 +1,19 @@
...
+  for={{this.inputId}}
+>
+  {{@field.label}}
+</AuLabel>
+
+{{#if @field.description}}
+  <p class="au-u-light au-u-margin-bottom-tiny">{{@field.description}}</p>
+{{/if}}
+
+<RdfaEditorContainer
```
Windvis, line 13: I would inline the component in the rich text component it doesn't provide that many benefits of splitting it up I think. Alternatively you could add it to a subfolder so it's clear that it's a "private" component? `<RdfInputFields::RichText::Editor>`

addon/components/rdfa-editor-container.hbs
```diff
@@ -0,0 +1,45 @@
...
+                <div class="au-c-scanner__text">
+                  <AuLoader @padding="small"/>
+                  <AuHelpText @size="large">{{if @busyText @busyText
+                                                 (t "rdfaEditorContainer.defaultBusyText") }}
+                  </AuHelpText>
+                </div>
+                <span class="au-c-scanner__bar"></span>
+              </div>
+              <div class="say-editor__inner say-content">
+                {{html-safe this.editor.htmlContent}}
```
Windvis, line 18: This helper doesn't exist in the addon so it crashes if you set `@busy` to `true`.

addon/components/rdfa-editor-container.hbs
```diff
@@ -0,0 +1,45 @@
...
+  </AuBodyContainer>
+{{else}}
+  <AuBodyContainer 
+    vocab="{{this.vocab}}" 
+    {{did-insert this.setPrefix}}
+    typeof="{{@typeOfWrappingDiv}}" 
+    property="{{@property}}" 
+    resource="#"
+  >
+    <div {{on "focusout" @onBlur}}>
```
Windvis, line 34: This seems very strange to me. Doesn't the editor provide callbacks that we can use instead?

addon/components/rdfa-editor-container.hbs
```diff
@@ -0,0 +1,45 @@
...
+      <Rdfa::RdfaEditor
+        class="au-c-rdfa-editor"
```
Windvis, line 36: It seems this won't get passed through since the rdfa-editor component doesn't use `...attributes`.

addon/components/rdfa-editor-container.hbs
```diff
@@ -0,0 +1,45 @@
...
+    </div>
+  </AuBodyContainer>
+{{else}}
+  <AuBodyContainer 
+    vocab="{{this.vocab}}" 
+    {{did-insert this.setPrefix}}
+    typeof="{{@typeOfWrappingDiv}}" 
+    property="{{@property}}" 
+    resource="#"
+  >
```
Windvis, line 33: Is this body container needed? Seems like a strange component to use for an inline editor?

addon/components/rdfa-editor-container.hbs
```diff
@@ -0,0 +1,45 @@
...
+            </div>
+          </div>
+        </div>
+      </div>
+    </div>
+  </AuBodyContainer>
+{{else}}
+  <AuBodyContainer 
+    vocab="{{this.vocab}}" 
+    {{did-insert this.setPrefix}}
```
Windvis, line 29: The addon doesn't list `@ember/render-modifiers` as a dependency so at the moment this only works by accident since another addon brings it in. 

Are these rdfa attributes needed for the editor to work? I don't see it documented in the editor readme so if it's not needed I would just delete the wrapper altogether?

Maybe @cecemel has a different opinion.

addon/components/rdfa-editor-container.js
```diff
@@ -0,0 +1,83 @@
...
+  get editorOptions() {
+    return (
+      this.args.editorOptions ?? {
+        showToggleRdfaAnnotations: Boolean(this.args.showToggleRdfaAnnotations),
+        showInsertButton: false,
+        showRdfa: true,
+        showRdfaHighlight: true,
+        showRdfaHover: true,
+        showPaper: true,
+        showSidebar: true,
```
Windvis, line 17: I think some of these can be set to false to make it more suitable for the "inline" editor use-case?

Windvis (changes requested):
I left some comments, but I believe the idea was to only add this field to the Loket frontend for now? I would have expected this PR to be made there instead 😄.

addon/components/rdf-input-fields/rich-text.hbs
lagartoverde: The thing is that we are going to use it for other components, like now with the rdfa-editor component that includes plugins

addon/components/rdfa-editor-container.hbs
lagartoverde: I used it to mimic the setup in GN I guess we can do better when we customize the styling

package.json
```diff
@@ -27,8 +27,12 @@
     "test:ember-compatibility": "ember try:each"
   },
   "dependencies": {
+    "@codemirror/basic-setup": "^0.20.0",
+    "@codemirror/lang-html": "^0.20.0",
+    "@codemirror/lang-xml": "^0.20.0",
```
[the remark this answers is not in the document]
lagartoverde: I asked arne and he responded `ye we havent figured out yet how to make them actualy optional` so I guess help is needed there

addon/components/rdfa-editor-container.hbs
lagartoverde: I wanted to use it, as it would be useful for showing when loading but I didn't find a use for it yet, so I guess I will remove it

package.json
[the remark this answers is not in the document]
Windvis: I _think_ it should be possible by using the `dependencySatisfies` and `importSync` [macros](https://www.npmjs.com/package/@embroider/macros). So the debug component could try to import them and throw an assertion if they are not installed.

Windvis:
Another thing I noticed is that the component doesn't have a "read only" version yet. Most forms have a read only state (which can be mimicked with the toggle in to top bar of the test app) which should prevent users from editing the contents.

Each field component receives a `@show` attribute which can be used to conditionally alter the behavior.

addon/components/rdfa-editor-container.hbs
lagartoverde: Depends on what we want to do respect to rdfa with the rich text editor, the point of this prefixes + vocab + everything is to provide a wrapper around the content, for example in GN the wrapper is the behandelingVanAgendapunt, and all the editor content is inside that.
This wrapper sets the context and the prefixes

Windvis: The way I understood it is that rdfa-editor is just an implementation detail for the rich-text field. Users should just be able to edit it in a user-friendly way. So I would make it highly opinionated for that use-case and not think about the rdfa editing for now.

So I would just disable everything that we don't need. Things like the sidebar, "paper", prefix, etc. We can always refactor things later when we actually need rdfa editing and there is obvious logic duplication going on.

It would make adding styles for this a lot easier as well. We could probably just borrow the setup that's used for the AuTextarea component but with a toolbar at the top. I feel we can even manage a first iteration without any designer support 😄.

cecemel: I don't think for the rich-text-editor this should be needed. In theory we should be able to swap it with any other rich-text-editor

package.json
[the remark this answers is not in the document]
nvdk: yeah it's an ember thing, there's no tree shaking on unused ember components and it will complain that the debug component can't import these even if you don't use the component. this should be fixed when we can move to embroider in ember-rdfa-editor I think, but we're pretty far from that.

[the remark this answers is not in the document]
Windvis: No, I think it should still be possible in classic builds with the Embroider macros mentioned above. The idea is that the debug component uses importSync to import the dependencies instead of regular imports so ember-auto-import won't complain at build time about the missing dependency. Consumers using the debug component will then have to ensure the dependencies are installed, or they will see the errors at runtime when using the component (although it's probably better to throw a custom assertion with proper guidance). It works in my head at least, but I haven't actually tried it so I might be missing important details 😄.

Another option for classic builds would be to use the `treeFor` hooks in the addon to strip out the debug component if a certain build flag is set (or not set depending on what makes the most sense as a default), but that's less ideal since it won't work if you want to convert the addon to the v2 addon format (, so it's guaranteed tech debt).

Yet another option could be to move the debug component to a dedicated addon so users who want it can simply install it?

cecemel:
Closing here, because solved in https://github.com/lblod/frontend-loket/pull/127
thanks for all the explorative work!

PR #79, opened 2022-06-02
Feature/updated form model

cecemel:


cecemel:
@Windvis If you have time best to have a look. 
Ideally make difference between major (i.e. blocking) for a release and minor (may be picked up later)
thanks!

app/models/listing.js
```diff
@@ -0,0 +1 @@
+export { default } from '@lblod/ember-submission-form-fields/models/listing';
```
Windvis, line 1: Nitpick, can be done in the future since this pattern was already used before: 

I don't think the "models" need to be reexported from the app folder. These are only used internally by the addon I think? And even if apps would use them directly they would probably still import them from `@lblod/embersubmission-form-fields/models/listing.js` instead of `app-name/models/listing.js`?

tests/dummy/app/templates/form.hbs
```diff
@@ -3,4 +3,33 @@
...
+<AuHeading
+  @level={{1}}
+  @skin={{3}}
+>
+  Data
+</AuHeading>
+<AuTable
+  @size="large"
+>
+  <:title>Test</:title>
```
Windvis, line 18: I would remove the heading and change it to this, or remove the title block and only keep the heading.

```suggestion
  <:title>Data</:title>
```

tests/dummy/app/templates/form.hbs
```diff
@@ -3,4 +3,33 @@
...
+  <:header>
+  <tr>
+    <th>Subject</th>
+    <th>Predicate</th>
+    <th>Object</th>
+  </tr>
+  </:header>
+  <:body>
+  {{#each this.datasetTriples as |triple|}}
```
Windvis, line 27: Nice addition! although it does get very long on some forms, so we can probably improve the rendering in the future.

package.json
```diff
@@ -29,7 +29,7 @@
   "dependencies": {
     "@embroider/macros": "^1.6.0",
     "@embroider/util": "^1.6.0",
-    "@lblod/submission-form-helpers": "^1.3.0",
+    "@lblod/submission-form-helpers": "https://github.com/lblod/submission-form-helpers#262d0c1511f4d3502eb75f73f02b43a6aa11b3d0",
```
Windvis, line 32: I guess we can merge the submission-form-helpers PR, release a new version and use that here before merging this one?

Windvis (approved):

PR #84, opened 2022-07-12
Replace browser-rdflib with rdflib v2

Windvis:
rdflib v2 is browser compatible so we no longer need the browser-rdflib package (which is based on rdflib v1).

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,9 +29,8 @@
   "dependencies": {
     "@embroider/macros": "^1.6.0",
     "@embroider/util": "^1.6.0",
-    "@lblod/submission-form-helpers": "^1.4.1",
+    "@lblod/submission-form-helpers": "^2.0.1",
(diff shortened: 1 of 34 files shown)


package.json
```diff
@@ -29,9 +29,8 @@
   "dependencies": {
     "@embroider/macros": "^1.6.0",
     "@embroider/util": "^1.6.0",
-    "@lblod/submission-form-helpers": "^1.4.1",
+    "@lblod/submission-form-helpers": "git@github.com:lblod/submission-form-helpers.git#9a84dc13cb36c902d28f9e35b3822e8e98781239",
```
Windvis, line 32: Using a specific commit until the branch is merged and released. Both `@lblod/submission-form-helpers` and `forking-store` use rdflib internally so we use the branch where the browser-rdflib to rdflib 2 conversion happened.

PR #94, opened 2022-08-11
Fallback to an empty string if a literal value is null or undefined

Windvis:
Rdflib literals can only have string or number values and will throw an error if `null` is passed.

### DIFF
--- a/addon/components/rdf-input-fields/simple-value-input-field.js
+++ b/addon/components/rdf-input-fields/simple-value-input-field.js
@@ -67,7 +67,7 @@ export default class SimpleValueInputFieldComponent extends InputFieldComponent 
     } else {
       literalOrValue = this.nodeValue?.copy();
       if (literalOrValue) {
-        literalOrValue.value = value;
+        literalOrValue.value = value || '';
       } else {
         literalOrValue = value;
       }


addon/components/rdf-input-fields/simple-value-input-field.js
```diff
@@ -67,7 +67,7 @@ export default class SimpleValueInputFieldComponent extends InputFieldComponent
     } else {
       literalOrValue = this.nodeValue?.copy();
       if (literalOrValue) {
-        literalOrValue.value = value;
+        literalOrValue.value = value || '';
```
Windvis, line 70: This isn't the right fix. This will convert all falsy values to an empty string which we don't want.

 But it seems the checkbox field has a similar issue where it sets a boolean value with also throws an exception so we need to something else to fix it.

cecemel: Yeah falsy is not specific enough. Should test for null and undefined?

-> not sure what the types of values we expect? 
We either expect a string, or a named node right?

Windvis:
This PR is superceded by #95 since that fixes the problem in a better way.

PR #95, opened 2022-08-17
Only use the lang string logic if the predicate is set

Windvis:
This refactors the code to only use the new lang string behavior if the `form:language` predicate is set. The introduction of the lang-string feature broke other components in unexpected ways. By hiding that logic behind a conditional statement we restore the old behavior for non-lang-string fields.

### DIFF
--- a/addon/components/rdf-input-fields/simple-value-input-field.js
+++ b/addon/components/rdf-input-fields/simple-value-input-field.js
@@ -14,84 +14,62 @@ export default class SimpleValueInputFieldComponent extends InputFieldComponent 
     this.loadProvidedValue();
   }
 
+  get isLangStringField() {
+    return Boolean(this.args.field.language);
+  }
+
   loadProvidedValue() {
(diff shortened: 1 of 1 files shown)


addon/components/rdf-input-fields/simple-value-input-field.js
```diff
@@ -14,86 +14,64 @@ export default class SimpleValueInputFieldComponent extends InputFieldComponent
...
-        // Note: it uncovers wonky side effects,
-        //  in case we have two different literals for the same predicate.
-        //  Updating one of them might result in strange rendering behaviour.
-        // This was the case in the previous version too, and might need re-thinking. (TODO)
-        // If not clear, try it in the dummy-app
-        if (!literal) {
-          literal = literals[0];
-        }
-      }
+    if (this.isLangStringField) {
```
Windvis, line 24: This restores the old behavior if no lang string predicate is set, which should fix all the regressions we introduced.

PR #96, opened 2022-08-17
Make the PropertyGroup title level and skin configurable

Windvis:
This allows us to override the default level 1 / skin 2 combo. Overriding the level also affects the levels of nested PropertyGroups, so simply setting this on the root PropertyGroup should work. The "skin" can also be adjusted in case a different visual style is needed for the design.

`skin` and `level` support the same values as the [AuHeading component](https://appuniversum.github.io/ember-appuniversum/?path=/story/components-content-auheading--component).

Allowing to change the level means we can fit in a form on a page which already has h* elements, and still output a valid header hierarchy.

### DIFF
--- a/addon/components/property-group.hbs
+++ b/addon/components/property-group.hbs
@@ -1,13 +1,15 @@
 {{#if this.children}}
   <div ...attributes>
-    {{#if @group.name}}
+    {{#if this.showTitleBlock}}
       <div class="au-u-margin-bottom-small">
-        <AuHeading @level={{this.titleLevel}} @skin={{this.titleSkin}}>
-          {{@group.name}}
-        </AuHeading>
+        {{#if @group.name}}
+          <AuHeading @level={{this.titleLevel}} @skin={{this.titleSkin}}>
+            {{@group.name}}
+          </AuHeading>
+        {{/if}}
         {{#if @group.help}}
           {{! template-lint-disable no-triple-curlies}}
-          <p class="au-u-light au-u-margin-top-xsmall">{{{@group.help}}}</p>
+          <p class="au-u-light">{{{@group.help}}}</p>
         {{/if}}
         {{#each this.errors as |error|}}
           <AuAlert class="au-u-margin-bottom-small au-u-margin-top-small"
--- a/addon/models/property-group.js
+++ b/addon/models/property-group.js
@@ -1,5 +1,4 @@
 import { tracked } from '@glimmer/tracking';
-
 import { FORM, SHACL } from '@lblod/submission-form-helpers';
 
 export const PROPERTY_GROUP_DISPLAY_TYPE =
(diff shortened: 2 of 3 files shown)


addon/components/property-group.hbs
```diff
@@ -1,13 +1,19 @@
 {{#if this.children}}
   <div ...attributes>
-    {{#if @group.name}}
+    {{#if this.showTitleBlock}}
```
Windvis, line 3: This change also allows us to show help text even if we don't want to show a title. This should remove the need to hack around it by setting the title to an empty string (`sh:name ""`)

addon/models/property-group.js
```diff
@@ -25,6 +28,12 @@ export default class PropertyGroupModel {
       formGraph
     );
     this.rdflibHelp = store.any(uri, FORM('help'), undefined, formGraph);
+
+    let rdfLevel = store.any(uri, EXT('level'), undefined, formGraph);
```
Windvis, line 32: I've used an `ext` prefix for now since I wasn't sure which predicate uris to use. Suggestions?

Should we solve this in another way?

Windvis:
It seems the `SubForm`s [use a similar setup](https://github.com/lblod/ember-submission-form-fields/blob/9c4cd929ac6ecdee08b0f0fbba4014180e4d7e7f/addon/components/sub-form.hbs#L1-L7) so we might also want this change there?

addon/models/property-group.js
Windvis: Maybe we should pass the data through `form:options` instead?

FangioD:
I tested in dummy this change, and it works fine, but @Windvis could you explain the semantic difference betweend `ext:level` and `ext:skin`?  The skin is dominant over level, but what does it mean if you both add them?

Windvis:
The level is the thing that outputs the actual `h*` element. By default it starts with `h1` and it increases for each level of property group nesting. It also influences the skin if the skin options isn't set. The skin is simply a visual cue. You can choose whatever you want for it, which can be used to match the design. So in your example it will output a `h4` element which is styled as a `h3`.

TL;DR use `level` to set a top level header element which is appropriate for the page. Only use the skin if the styling doesn't match the design.

addon/models/property-group.js
Windvis: I went with the form:options version.

Windvis:
@FangioD I've changed the way the configuration is passed. Now it uses the `form:options` predicate, similar to how we already do it for some fields.

For example:
```ttl
fields:group-id a form:PropertyGroup;
    form:options """{ "level": 2, "skin": 6 }""".
```

Windvis:
Released as [v2.1.0](https://github.com/lblod/ember-submission-form-fields/blob/master/CHANGELOG.md#v210-2022-08-30)

PR #98, opened 2022-08-22
Improve the read-only mode of the Listing and SubForm components

Asergey91:


addon/components/listing.hbs
```diff
@@ -17,16 +17,18 @@
...
-        @icon="plus"
-        @iconAlignment="left"
-        class="au-u-margin-bottom-tiny"
-        @disabled={{this.creationFormData}}
-        {{on "click" this.createEntry}}
-      >
-        {{@listing.addLabel}}
-      </AuButton>
-    </div>
+    {{#unless @show}}
```
Windvis, line 20: I think moving this into the `canAdd` getter might be better?

Windvis (changes requested):
It seems the subform's `canRemove` logic doesn't handle the `@show` state yet either so you can still remove things in "readonly" mode at the moment. Could you maybe add that as part of this PR? 🙏

Windvis (approved):

PR #102, opened 2022-10-24
Heading component, independently of PropertyGroup

FangioD:
> Sometimes forms asks for some helping text, which we currently (in some cases) handled by adding a skin/level to the property-group, but that is actually a bit overkill.

Added a paragraph component, this component acts like a label without the need for any inputtype. Example is given in dummy-app in the basic-fields form.


addon/components/rdf-input-fields/paragraph.hbs
```diff
@@ -0,0 +1,3 @@
+<AuHeading @skin={{this.titleSkin}}>
```
Windvis, line 1: I wonder if we should create a Heading field component instead which would simply be the `AuHeading` component in field form, similar to how it works in the PropertyGroup. Instead of hardcoding the skin here.

That might be more flexible and potentially allows us to drop the title feature of the PropertyGroup in the future as well.

addon/components/rdf-input-fields/paragraph.js
```diff
@@ -0,0 +1,7 @@
+import Component from '@glimmer/component';
+
+export default class RdfParagraphComponent extends Component {
```
Windvis, line 3: FYI if you run into the [no-empty-glimmer-component-classes](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-empty-glimmer-component-classes.md#no-empty-glimmer-component-classes) linting rule you can simply delete the .js file.

addon/components/rdf-input-fields/paragraph.hbs
FangioD: I tried to get it like the PropertyGroup component, instead of returning "6", however it did not seem to work.

```  
   get titleSkin() {
    return this.args.group.options.skin || `${this.level + 1}`;
  }
``` 

I added `form:options` with skin like the PropertyGroup, but
`this.args.group` is undefined.

Windvis: Yea, `@group` is only passed into property groups. It's `@field` for regular form fields. So `this.args.field.options.skin` should work I think.

addon/components/rdf-input-fields/heading.hbs
```diff
@@ -0,0 +1,3 @@
+<AuHeading @skin={{this.titleSkin}}>
```
Windvis, line 1: I would also pass the `@level` argument so it can be adjusted. Like here: https://github.com/lblod/ember-submission-form-fields/blob/e36be75f1479c5c4a8c792cf69250752fed6b601/addon/components/property-group.hbs#L6

addon/components/rdf-input-fields/heading.js
```diff
@@ -0,0 +1,7 @@
+import Component from '@glimmer/component';
+
+export default class RdfHeadingComponent extends Component {
+  get titleSkin() {
```
Windvis, line 4: Nitpick: I would rename this to `skin` since the title part is implicit in this component. (same for the level).

addon/components/rdf-input-fields/heading.js
```diff
@@ -0,0 +1,7 @@
...
+    return this.args.field.options.skin || '6';
```
Windvis, line 5: I would use the same defaults as the AuHeading component so just return the option value here maybe? That way it will use the AuHeading default value if it's undefined. It requires adding the config for your use-case, but I think it's a better long term default value.

FangioD: @Windvis if you only add `@level` argument, nothing will change to the skin, shouldn't the `@skin` argument be related to the level if skin is not passed?

Windvis: Yea, that's a valid point. That's more a question for Appuniversum though 😄.

Windvis (approved):

tests/dummy/public/test-forms/basic-fields/form.ttl
```diff
@@ -34,6 +34,28 @@ fields:8e24d707-0e29-45b5-9bbf-a39e4fdb2c11 a form:PropertyGroup;
...
+    form:help """ This is a 
+    <a href="https://lokaalklimaatpact.be/formulieren.html" target="_blank">link</a>. <br>
+    <ul  style="list-style-type:disc; margin-left: 30px;">
+  <li>Coffee</li>
+  <li>Tea</li>
+  <li>Milk</li>
+</ul>
+    """ ;
+    sh:order 190;
+    form:options """{ "level": 1, "skin": 6}""";
```
Windvis, line 52: oh, right, the values should be strings, or at least the level one. Appuniversum uses strict equals checks for the level so it will always default to `1` if it doesn't match  any of the strings.

https://github.com/appuniversum/ember-appuniversum/blob/development/addon/components/au-heading.hbs

It's a bit of a strange API I guess, but that's the way it is implemented at the moment 😄. `skin` is fine since it does a `==` check so js converts the number into a string first. But it might be nice to still convert them to strings anyway. I'll push the change since I have it open anyways.

Windvis:
Released as [v2.2.0](https://github.com/lblod/ember-submission-form-fields/blob/master/CHANGELOG.md#v220-2022-10-25)

PR #104, opened 2022-11-15
Add a "table-mode" to the listing component

Windvis:
This change allows us to display listings as a table instead of a list of blocks.

I added an example table that mimics the [Figma design](https://www.figma.com/file/DsCBi6M992vosTZAQ1LyzL/formbuilder?node-id=0%3A1&t=LBy2tDq6jeVvuXdr-0) and looks like this:

image

### DIFF
--- a/addon/components/listing.hbs
+++ b/addon/components/listing.hbs
@@ -7,7 +7,7 @@
   @level={{@level}}
   @removeEntry={{this.removeEntry}}
   @canAdd={{this.canAdd}}
-  @canRemove={{@listing.canRemove}}
+  @canRemove={{this.canRemove}}
   @createEntry={{this.createEntry}}
   @listing={{@listing}}
   @sourceNode={{@sourceNode}}
--- a/addon/components/listing.hbs
+++ b/addon/components/listing.hbs
@@ -1,32 +1,14 @@
-<div class="au-o-grid">
-  <div class="au-o-grid__item au-u-4-5">
-    {{#each this.subForms as |subForm index|}}
-      <SubForm
-        @subForm={{subForm}}
-        @formStore={{@formStore}}
-        @graphs={{@graphs}}
-        @sourceNode={{subForm.sourceNode}}
-        @forceShowErrors={{@forceShowErrors}}
-        @cacheConditionals={{@cacheConditionals}}
-        @last={{eq index (sub this.subForms.length 1)}}
-        @show={{@show}}
-        @level={{@level}}
-        @onRemoveEntry={{this.removeEntry}}
-        @canRemove={{@listing.canRemove}}
-      />
-    {{/each}}
-  </div>
-  {{#if this.canAdd}}
-    <div class="au-o-grid__item au-u-1-5">
-      <AuButton
-        @icon="plus"
-        @iconAlignment="left"
-        class="au-u-margin-bottom-tiny"
-        @disabled={{this.creationFormData}}
-        {{on "click" this.createEntry}}
-      >
-        {{@listing.addLabel}}
-      </AuButton>
-    </div>
-  {{/if}}
-</div>
+<this.ListingDisplayMode
+  @subForms={{this.subForms}}
+  @formStore={{@formStore}}
+  @graphs={{@graphs}}
+  @forceShowErrors={{@forceShowErrors}}
+  @show={{@show}}
+  @level={{@level}}
+  @removeEntry={{this.removeEntry}}
+  @canAdd={{this.canAdd}}
+  @canRemove={{@listing.canRemove}}
+  @createEntry={{this.createEntry}}
+  @listing={{@listing}}
+  @sourceNode={{@sourceNode}}
+/>
--- a/addon/components/listing/list.hbs
+++ b/addon/components/listing/list.hbs
@@ -12,7 +12,7 @@
         @show={{@show}}
         @level={{@level}}
         @onRemoveEntry={{@removeEntry}}
-        @canRemove={{@listing.canRemove}}
+        @canRemove={{@canRemove}}
       />
     {{/each}}
   </div>
--- a/addon/components/listing/list.hbs
+++ b/addon/components/listing/list.hbs
@@ -1,0 +1,32 @@
+<div class="au-o-grid">
+  <div class="au-o-grid__item au-u-4-5">
+    {{#each @subForms as |subForm index|}}
+      <SubForm
+        @subForm={{subForm}}
+        @formStore={{@formStore}}
+        @graphs={{@graphs}}
+        @sourceNode={{subForm.sourceNode}}
+        @forceShowErrors={{@forceShowErrors}}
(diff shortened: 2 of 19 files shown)


addon/components/listing.hbs
```diff
@@ -1,32 +1,98 @@
...
-        @show={{@show}}
-        @level={{@level}}
-        @onRemoveEntry={{this.removeEntry}}
-        @canRemove={{@listing.canRemove}}
-      />
-    {{/each}}
-  </div>
-  {{#if this.canAdd}}
-    <div class="au-o-grid__item au-u-1-5">
+{{#if @listing.isTable}}
```
Windvis, line 1: I've used big conditional blocks for now since that allowed me to iterate faster, but it might be worthwhile to work with different components instead?

addon/components/listing.hbs
```diff
@@ -1,32 +1,98 @@
...
+  <div>
+    <AuTable class="listing-table">
+      <:title>
+        {{! TODO: read this from the graph directly }}
+        {{this.subForms.firstObject.itemLabel}}
```
Windvis, line 6: I've used the title of the first subForm for now. Maybe it's better to read this from the form graph instead?

addon/components/listing.hbs
```diff
@@ -1,32 +1,98 @@
...
+      </:title>
+      <:header>
+        {{!TODO this can probably be done without the subforms setup by simply retrieving the labels from the graph in some way?}}
+        {{#let this.subForms.firstObject as |subForm|}}
```
Windvis, line 10: We need a list of al the labels of the form fields that belong to this listing. I've used the SubForm component as a hacky solution for now, but there must be better ways.

addon/components/listing.hbs
```diff
@@ -1,32 +1,98 @@
...
+              @canRemove={{@listing.canRemove}}
+              @isRow={{true}}
+              @isTableHead={{true}}
+            />
+          {{/if}}
+        {{/let}}
+      </:header>
+      <:body>
+        {{! TODO: Show something if the table is empty }}
+        {{#each this.subForms as |subForm index|}}
```
Windvis, line 32: I've reused the subForm system of regular listings since that keeps the form.ttl the same. Would it be better to go with a different setup specifically made for tables instead?

addon/components/rdf-input-fields/concept-scheme-selector.hbs
```diff
@@ -1,10 +1,12 @@
-<AuLabel
-  @error={{this.hasErrors}}
-  @required={{this.isRequired}}
-  for={{this.inputId}}
->
-  {{@field.label}}
-</AuLabel>
+{{#unless @isTable}}
```
Windvis, line 1: At the moment I'm reusing the existing form fields by passing an extra argument into it. They can then decide how they should be displayed if rendered in a table. I've only done it for a couple of components. Not sure if it would be better to create dedicated components or not.

Windvis: The argument name is also up-for-debate but didn't want to lose to much time on it.

tests/dummy/public/test-forms/tables/form.ttl
```diff
@@ -0,0 +1,98 @@
...
+#  listing-scope: administrativeUnit
+##########################################################
+ext:administrativeUnitListingS a form:Scope;
+  sh:path ext:administrativeUnit.
+
+##########################################################
+#  Listing: Administrative unit
+##########################################################
+ext:administrativeUnitsL a form:Listing;
+  a form:ListingTable; #TODO
```
Windvis, line 33: At the moment this is the only change compared to the regular listing / sub form setup. Adding this class switches the display to the table version.

Is this the way to go or are there better options? We could also drop the form:Listing class and create whole new components instead of reusing things.

addon/components/listing.hbs
cecemel: some kind of title field maybe?

tests/dummy/public/test-forms/tables/form.ttl
cecemel: Well the good thing here, for the backend, it won't change a thing. full transparent. 
The subform however, when in table mode, seems indeed also affected and worth a component on it's own. So perhaps double type here, and split table listing and table subform into really their own components no?

We need further discussion for sure.

addon/components/rdf-input-fields/concept-scheme-selector.hbs
cecemel: I am also inclined that 's it is going to be messy once we extend features.

addon/components/listing.hbs
cecemel: yeah the more i look at it the more it feels like this no?

addon/components/listing.hbs
cecemel: This is an argument for an extra headers component defined in the form.ttl (as long as it is transparent for the backend)

cecemel (changes requested):
At first glance, I liked how the table got plugged into the base building blocks of the form definition and code. But the longer I looked at it, the more scared I got about further extensibility. And it seems your questions point toward this, too, no?

From the perspective of the backend, we should aim for full transparency. 
In the front, it felt a bit too many "if else's" to be extensible in the future. 
Also, both subform and listing, in table mode, seem closely tied to each other, which again points towards separate components, where we try to re-use the whole adding/removing entries implementation no?

Anyway, further discussion is needed, but also lovely progress. Thanks!

addon/components/listing/list.hbs
```diff
@@ -0,0 +1,32 @@
+<div class="au-o-grid">
```
Windvis, line 1: This is the original "listing" template.

addon/components/listing/table.hbs
```diff
@@ -0,0 +1,65 @@
+<div ...attributes>
+  <AuTable class="listing-table">
+    {{! TODO: Named blocks can't be rendered conditionally yet, so we would have to duplicate the component and all the other blocks as a workaround.
+        Since the title is only a "nice to have" we'll disable it for now until a better solution is found.
+    }}
+    {{!--
+    <:title>
+      {{this.title}}
+    </:title>
+    --}}
```
Windvis, line 10: I've disabled the table "title" for now since it looks weird if it isn't set and we would need to duplicate most of the template if we want to render it conditionally.

addon/components/listing/table.hbs
```diff
@@ -0,0 +1,65 @@
...
+              @size="small"
+              class="au-u-margin-top-none au-u-regular"
+            >
+              {{column.help}}
+            </AuHelpText>
+          {{/if}}
+        </th>
+      {{/each}}
+
+      {{#if @canRemove}}
```
Windvis, line 28: TODO: We need to disable this if the table is in read-only mode as well.

cecemel (approved):
great thanks!

PR #105, opened 2022-11-17
Remove the custom date picker localization

Windvis:
ember-appuniversum v1.10.0 added an option to localize the date picker in Dutch. This allows us to drop our own custom localization code.


ember-cli-build.js
```diff
@@ -5,6 +5,9 @@ const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 module.exports = function (defaults) {
   const customBuildConfig = {
     // Add options here
+    '@appuniversum/ember-appuniversum': {
+      dutchDatePickerLocalization: true,
+    },
```
Windvis, line 10: This only affects the test app so consuming apps need to add that config as well. We _could_ do it magically, but I think it's better if the app defines that config.

PR #107, opened 2022-12-01
Resolve "mutation after consumption" deprecations

Windvis:
Some components trigger the "X was updated but it was already used" deprecation when data is loaded on component creation. This is one of the final blockers to make the addon v4 ready.


Windvis:
I've added some dummy "persisted" data to the "basic components" form to test the data loading and I solved all cases I found. Hopefully that's all of them.

PR #120, opened 2023-02-17
Persist the order of listing entries

Windvis:
This adds an `sh:order` predicate to all sub forms so we can render in a consistent order.

### DIFF
--- a/addon/components/listing.js
+++ b/addon/components/listing.js
@@ -6,11 +6,14 @@ import {
   generatorsForNode,
   triplesForGenerator,
   triplesForScope,
+  SHACL,
 } from '@lblod/submission-form-helpers';
 import { getSubFormsForNode } from '../utils/model-factory';
 import { next } from '@ember/runloop';
 import ListingList from '@lblod/ember-submission-form-fields/components/listing/list';
 import ListingTable from '@lblod/ember-submission-form-fields/components/listing/table';
+
+const ORDER = SHACL('order');
 
 export default class ListingComponent extends Component {
   @tracked subForms = A();
@@ -54,6 +57,23 @@ export default class ListingComponent extends Component {
     return this.listing.isTable ? ListingTable : ListingList;
   }
 
+  get highestOrderValue() {
+    let highestOrderValue = 0;
+
+    if (this.subForms.length) {
+      let lastSubForm = this.subForms[this.subForms.length - 1];
+
+      highestOrderValue =
+        getOrder(
+          lastSubForm.sourceNode,
+          this.formStore,
+          this.graphs.sourceGraph
+        ) || 0;
+    }
+
+    return highestOrderValue;
+  }
+
   constructor() {
     super(...arguments);
     this.updateScope();
@@ -63,7 +83,7 @@ export default class ListingComponent extends Component {
     //  'You attempted to update `subForms` on `ListingComponent`, but it had already been used previously in the same computation. '
     // Why is it triggered? this.subForms is tracked, is rendered before the method below (which has side effects, i.e. updates this.subForms) ran.
     next(this, () => {
-      this.renderSubForms();
+      this.renderSubForms({ orderSubForms: true });
     });
   }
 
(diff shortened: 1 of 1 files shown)


addon/components/listing.js
```diff
@@ -6,12 +6,15 @@ import {
...
   triplesForGenerator,
   triplesForScope,
+  SHACL,
 } from '@lblod/submission-form-helpers';
 import { getSubFormsForNode } from '../utils/model-factory';
 import { next } from '@ember/runloop';
 import ListingList from '@lblod/ember-submission-form-fields/components/listing/list';
 import ListingTable from '@lblod/ember-submission-form-fields/components/listing/table';
 
+const ORDER = SHACL('order');
```
Windvis, line 16: I'm not sure this is the correct predicate. [`qb:order`](https://www.w3.org/TR/vocab-data-cube/#ref_qb_order) also seems to be a thing, and we use that for a [similar use-case in loket](https://github.com/lblod/frontend-loket/blob/0c5c628a4120502fd016fd468f2197b5e968866e/app/components/rdf-form-fields/climate-subsidy-costs-table.js#L168).

addon/components/listing.js
```diff
@@ -63,7 +83,7 @@ export default class ListingComponent extends Component {
     //  'You attempted to update `subForms` on `ListingComponent`, but it had already been used previously in the same computation. '
     // Why is it triggered? this.subForms is tracked, is rendered before the method below (which has side effects, i.e. updates this.subForms) ran.
     next(this, () => {
-      this.renderSubForms();
+      this.renderSubForms({ orderSubForms: true });
```
Windvis, line 86: We only do the sorting when the data is initially loaded, after that the ordering can't get out of order, since we only add new forms after the already existing ones.

addon/components/listing.js
```diff
@@ -75,6 +95,12 @@ export default class ListingComponent extends Component {
       ...dataset.triples.map((t) => {
         return { ...t, graph: this.graphs.sourceGraph };
       }),
+      {
+        subject: dataset.sourceNodes[0], // TODO: find out why there can be more than one source node, if so, we should probably insert an order triple for each one?
```
Windvis, line 99: This works, I'm not completely sure what this does exactly and in which cases more than one sourceNode would be returned.

addon/components/listing.js
```diff
@@ -243,3 +290,38 @@ export default class ListingComponent extends Component {
     return sourceData;
   }
 }
+
+function ensureOrderTriplesForExistingData({
+  subForms,
+  formStore,
+  sourceGraph,
+}) {
+  let shouldAddOrderTriples = subForms.some((subForm) => {
```
Windvis, line 299: My assumption was that if any sub form doesn't have a `sh:order` non of them will, so we add new ones.

I didn't check for cases where _some_ subforms have an order, but not all of them. In that case I don't know what the best way forward is, but I don't think it will happen anyways.

Let me know if the logic needs to be changed.

Windvis: Maybe we shouldn't do any fallback logic here and assume that the backend does it for the older sub forms?

Windvis:
Unrelated to this, it seems that creating and removing a generated item doesn't remove the triples from the store, or so it seems in the test app. I also tested it without the changes of this PR and it still happens so it's an older issue it seems. Not sure if it worked when originally implemented.

addon/components/listing.js
cecemel: From memory: there is complexity when you create a node that has inverse relation, e.g. you can have multiple authors attached to the new book you created. This is a case we basically don't support ATM

addon/components/listing.js
cecemel: A lookup of the definition i think it is ok to use this.

cecemel:
> 

yeah. (FROM MEMORY) This is known issue or rather a more complicated case. The initial idea was flushing everything that has been created, but sometimes you might remove too much
see  https://github.com/lblod/ember-submission-form-fields/blob/5dfbf65e0ee61fd3457c701f0533ea617c9fe43a/addon/components/listing.js#L112

Windvis:
The Embroider build seems to fail due to changes in Ember Data. I'll resolve that in a separate PR.

PR #122, opened 2023-03-28
Add an option to display the row index in the listing table

Windvis:
This adds a new `form:showTableRowIndex` predicate which can be used to display the row number as an extra column in the front of the table.

image

`form:tableIndexLabel` can be used to set a specific string as the column header text.

### DIFF
--- a/addon/components/listing/table.hbs
+++ b/addon/components/listing/table.hbs
@@ -9,6 +9,9 @@
     </:title>
     --}}
     <:header>
+      {{#if this.showRowIndex}}
+        <th>{{! index }}</th>
+      {{/if}}
       {{#each this.tableHeaders as |column|}}
         <th>
           {{column.label}}
@@ -30,7 +33,7 @@
       {{/if}}
     </:header>
     <:body>
-      {{#each @subForms as |subForm|}}
+      {{#each @subForms as |subForm index|}}
         <this.ListingTableRow
           @subForm={{subForm}}
           @formStore={{@formStore}}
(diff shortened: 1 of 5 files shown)


tests/dummy/public/test-forms/tables/form.ttl
```diff
@@ -38,6 +38,7 @@ ext:administrativeUnitsL a form:Listing;
   form:addLabel "Voeg nieuw betrokken lokaal bestuur toe";
   form:canRemove true;
   form:removeLabel "Verwijder";
+  form:showTableRowIndex true;
```
Windvis, line 41: We could also add something like 

```ttl
form:tableRowIndexPattern "Stap {}";
```

Where we could then replace the `{}` (or whatever placeholder pattern we choose), to display something like `Stap 1` to the user. But that might be very niche and not worth the effort?

addon/components/listing/table.hbs
```diff
@@ -9,6 +9,9 @@
     </:title>
     --}}
     <:header>
+      {{#if this.showRowIndex}}
+        <th>{{! index }}</th>
```
Windvis, line 13: This needs to be configurable as well.

tests/dummy/public/test-forms/tables/form.ttl
Windvis: Since the column header can also be configured now which covers the same use case.

tests/dummy/public/test-forms/tables/form.ttl
```diff
@@ -38,6 +38,8 @@ ext:administrativeUnitsL a form:Listing;
   form:addLabel "Voeg nieuw betrokken lokaal bestuur toe";
   form:canRemove true;
   form:removeLabel "Verwijder";
+  form:tableIndexLabel "#";
+  form:showTableRowIndex true;
```
Windvis, line 42: @cecemel Any remarks about the predicate naming?

aatauil (approved):

PR #131, opened 2023-04-21
Add `sh:minCount` support to the listing component

Windvis:
Similar to `sh:maxCount`, this can be used to add a minimum requirement to the amount of listing items. If the data doesn't contain enough items we now pregenerate extra items until the minimum is reached.

This can be used to force users to enter a certain amount of data without having to use constraints.

### DIFF
--- a/addon/components/listing.js
+++ b/addon/components/listing.js
@@ -1,4 +1,3 @@
-import { A } from '@ember/array';
 import { action } from '@ember/object';
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
@@ -16,7 +15,7 @@ import ListingTable from '@lblod/ember-submission-form-fields/components/listing
 const ORDER = SHACL('order');
 
 export default class ListingComponent extends Component {
-  @tracked subForms = A();
+  @tracked subForms = [];
   scope = null;
 
   get formStore() {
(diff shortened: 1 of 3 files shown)


addon/components/listing.js
```diff
@@ -16,7 +16,7 @@ import ListingTable from '@lblod/ember-submission-form-fields/components/listing
 const ORDER = SHACL('order');
 
 export default class ListingComponent extends Component {
-  @tracked subForms = A();
+  @tracked subForms = [];
```
Windvis, line 19: I refactored this to a normal array already which solves the "'You attempted to update `subForms` on `ListingComponent`" issue (but we still run into others).

addon/components/listing.js
```diff
@@ -165,52 +219,33 @@ export default class ListingComponent extends Component {
     }
 
     // 2) calculate to be removed
-    const deletes = this.subForms.filter(
+    const deletes = currentSubForms.filter(
```
Windvis, line 222: I think this whole setup can be simplified some more, but didn't want to spend to much time on it now.

cecemel:
@Windvis I think it looks like it should do what intended to do. The full render component thing, we'll have to go later for it. But merge as is now.

PR #148, opened 2023-09-22
feat: creation collapsible fields

ZijneDwaasheid:
Refs: SFB-17


ZijneDwaasheid:
You can make a collapsible field by adding `form:canCollapse true` in the form.ttl in the formbuilder frontend. For example, when adding `form:canCollapse true` under `ext:propertyGroupL` would make the sections collapsible. The same could be done for Fields etc.

aatauil:
I will be closing this PR for now. The branch will continue to exist as this might be picked up later in same form or another

PR #150, opened 2023-09-28
Add a new reordering option to the listings

Windvis:
This adds a new `form:canChangeOrder` predicate that adds new buttons which can be used to move listing items up or down. To enable the feature simply set the predicate to `true`.

```ttl
ext:someListing a form:Listing ;
  # a form:ListingTable ; # Optional, if you want the table display mode
  # ... other listing predicates
  form:canChangeOrder true .
```

<details>
<summary>Listing version</summary>

image

</details>

<details>
<summary>(old) Listing version</summary>

image
</details>

<details>
<summary>Table version</summary>

image
</details>

### DIFF
--- a/addon/components/listing.js
+++ b/addon/components/listing.js
@@ -13,6 +13,10 @@ import ListingList from '@lblod/ember-submission-form-fields/components/listing/
 import ListingTable from '@lblod/ember-submission-form-fields/components/listing/table';
 
 const ORDER = SHACL('order');
+const DIRECTION = {
+  UP: 'up',
+  DOWN: 'down',
+};
 
 export default class ListingComponent extends Component {
   @tracked subForms = [];
@@ -54,6 +58,12 @@ export default class ListingComponent extends Component {
       : true;
 
     return this.listing.canRemove && !this.args.show && hasMinimumItems;
+  }
+
+  get canChangeOrder() {
+    return (
+      this.listing.canChangeOrder && !this.args.show && this.subForms.length > 1
+    );
   }
 
   get ListingDisplayMode() {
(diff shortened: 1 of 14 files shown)


addon/components/listing.js
```diff
@@ -146,6 +146,118 @@ export default class ListingComponent extends Component {
...
+        object: newOrder,
+        graph: this.graphs.sourceGraph,
+      },
+      {
+        subject: nextSubform.sourceNode,
+        predicate: ORDER,
+        object: currentOrder,
+        graph: this.graphs.sourceGraph,
+      },
+    ]);
```
Windvis, line 255: This will trigger the "observers" twice, once for each operation. Ideally that only happens once but that requires changes in the forking-store library.

addon/components/sub-form.hbs
```diff
@@ -9,8 +9,33 @@
             </AuHeading>
           {{/if}}
         </Group>
-        {{#if @canRemove}}
+        {{#if (or @canRemove @canMoveUp @canMoveDown)}}
           <Group>
+            <AuButtonGroup @inline={{(and @canMoveUp @canMoveDown)}}>
+              {{#if @canMoveUp}}
+                {{log @moveUp}}
```
Windvis, line 16:
```suggestion
```

addon/components/sub-form.hbs
```diff
@@ -9,8 +9,33 @@
             </AuHeading>
           {{/if}}
         </Group>
-        {{#if @canRemove}}
+        {{#if (or @canRemove @canMoveUp @canMoveDown)}}
           <Group>
+            <AuButtonGroup @inline={{(and @canMoveUp @canMoveDown)}}>
```
Windvis, line 14: Only implemented for the "new listing layout" I'm not sure if the old one needs it. Only LPDC uses that (with some nasty CSS hacks to force fields side-by-side).

tests/dummy/public/test-forms/sub-forms-focus-crud/form.ttl
```diff
@@ -39,6 +39,7 @@ ext:chaptersL a form:Listing;
   form:createGenerator ext:chapterGenerator;
   form:canRemove true;
   form:canAdd true;
+  form:canChangeOrder true;
```
Windvis, line 42:
```suggestion
  form:canReorder true;
```

might be better?

addon/components/sub-form.hbs
Windvis: Added it to the old layout as well.

addon/components/listing.js
Windvis: I added a TODO comment.

tests/dummy/public/test-forms/sub-forms-focus-crud/form.ttl
Windvis: We'll keep the current predicate unless someone has strong opinions.

addon/components/listing.js
```diff
@@ -146,6 +152,77 @@ export default class ListingComponent extends Component {
...
 
+  @action moveUp(subForm) {
+    this.changeItemOrder(subForm, true);
+  }
+
+  @action moveDown(subForm) {
+    this.changeItemOrder(subForm, false);
+  }
+
+  changeItemOrder(subform, shouldMoveUp) {
```
JonasVanHoof, line 163: >Not mandatory
Could we make an "enum" (constant) for this instead of a boolean?
```js
const DIRECTION = {
   UP: 'up',
   DOWN: 'down
} 
```

So it will become 
```js
this.changeItemOrder(subForm, DIRECTION.UP)
```

JonasVanHoof:
@Windvis 
It's not breaking when:
- making the predicate true
- making the predicate false
- not adding the predicate to the listing
- switching the items in the frontend up and down
- adding chapters with the same name and moving them up and down (tried this because I first tried moving the tables but that was not the scope of this and saw possibly something strange when switching items with the same selected item )(ignore this one just for info)

PR looks ok

addon/components/listing.js
Windvis: Yea, I figured it was overkill since there were only 2 options, but it makes it bit more readable so I added it anyways. 👍

Windvis:
> I first tried moving the tables but that was not the scope of this and saw possibly something strange when switching items with the same selected item

Sorry, I'm not sure what this means, can you elaborate a bit?

JonasVanHoof:
> > I first tried moving the tables but that was not the scope of this and saw possibly something strange when switching items with the same selected item
> 
> Sorry, I'm not sure what this means, can you elaborate a bit?

You can ignore this.. I cannot recreate the issue anymore atm :(  
image

It looked that when adding an item here with the same selected option that it could not be moved anymore.

Windvis:
Hmm, strange. It sounds like something might have thrown an exception somewhere which makes the app unresponsive. I couldn't replicate it either so I'll indeed ignore it for now. Thanks!

PR #154, opened 2023-10-18
Feat: fields with validations

JonasVanHoof:
## ID
/

 ## Description

While making it possible to add validations to fields in the `form-builder` and updating the wiki with the validation types and how to add them. I thought maybe its handy to have a page in the dummy app that shows the fields with the possible validations.

I added two starters to maybe start this an update it for all field validations:
image

 ## Type of change

 - [ ] Bug fix
 - [ ] New feature
 - [ ] Breaking change
 - [x] Other

 ## How to test

Run the dummy app and select the tab `basic form fields with validations`. Test all the fields displayed in that view.

 ## Links to other PR's

 - https://binnenland.atlassian.net/browse/SFB-39


JonasVanHoof:
Currently I have add these for default input & numerical input fields:
* is required
* exact value
* positive number
* phone number
* email

JonasVanHoof:
Leaving branch there for possible reference

PR #157, opened 2023-11-22
fix: order can be undefined

Rahien:
when using getOrder, e.g. to decide the order of a new field being added, the order can be undefined if the existing field didn't have an order. This causes getOrder to raise an exception. This sets the order to 0 if no order is defined.

### DIFF
--- a/addon/components/listing.js
+++ b/addon/components/listing.js
@@ -437,5 +437,5 @@ function hasOrder(subject, store, graph) {
 function getOrder(subject, store, graph) {
   let order = store.any(subject, ORDER, undefined, graph);
 
-  return parseInt(order.value);
+  return parseInt(order?.value) || 1;
 }


addon/components/listing.js
```diff
@@ -437,5 +437,5 @@ function hasOrder(subject, store, graph) {
 function getOrder(subject, store, graph) {
   let order = store.any(subject, ORDER, undefined, graph);
 
-  return parseInt(order.value);
+  return parseInt(order?.value) || 0;
```
Rahien, line 440: do you feel that 0 is a sane default or do you feel that undefined makes more sense?

JonasVanHoof: _haven't tested it on the ember submission form fields dummy app_

Should definitely return a number as I assume that it cannot handle the `null` or `undefined` as an order value.  Is zero the first order number? Never seen a field have order zero before so It needs to be `1`? 

Question: should the component be responsible to add the default value to the tripleStore?

Note: Maybe a strange thought but if you create a listing that has as default one field assigned. And the second field cannot get the order and falls back on the default than the second field will have the default order of `0` (or the `1` depends on what we end up to taking)

JonasVanHoof:
@Rahien Could you write a little `how to test` this in the PR description maybe than I can run the dummy app and quickly test this :D

Already thanks for checking it out!

addon/components/listing.js
Rahien: changed the order to 1 if undefined! i would not add anything to the triplestore in this case. Feels like a higher level thing.

Rahien:
> @Rahien Could you write a little `how to test` this in the PR description maybe than I can run the dummy app and quickly test this :D
> 
> Already thanks for checking it out!

So it turns out that it's really hard to reproduce this issue in the dummy app as there all the forms call ensureOrderTriplesForExistingData before asking for the order of their items. The problem occurs when the forms are dynamic as in the form builder: https://github.com/lblod/frontend-form-builder

When adding a listing there, you can see that it errors without this change but works with it.

Because of ember complaining about different versions of peer dependencies i had a hard time testing out my changes in this library in that app though. Here is the path that worked for me:

use yalc: https://github.com/wclr/yalc

in this branch of this repo: `yalc publish`

in the form builder repo: ``yalc add @lblod/ember-submission-form-fields`

you now have this version of the plugin in the form builder's frontend and the bug doesn't happen anymore when adding a field to a listing.

JonasVanHoof:
> > So it turns out that it's really hard to reproduce this issue in the dummy app as there all the forms call ensureOrderTriplesForExistingData before asking for the order of their items. The problem occurs when the forms are dynamic as in the form builder: https://github.com/lblod/frontend-form-builder
> 
> I'm a bit confused by this. It's been a while since I worked on this. The current implementation expects that all listing items have an order predicate. If (some) items don't have it, we generate new ones for all of them, as you already mentioned. But it seems you encountered a case where this doesn't happen? So the listing component is created, `ensureOrderTriplesForExistingData` runs, but after that there are still items without an order? What are the reproduction steps in the form builder app?
> 
> I think there might be an underlying issue here that we need to fix.

When generating a listing in the form builder one field is always pre-set and didn't have an order [here in form-builder](https://github.com/lblod/frontend-form-builder/commit/78caa863952a556a20dc58dbfdaa13ee382bfa45) i added an order for that field and now this is working accordingly. The next field added will have an order +1 from the first field. 

In the form-builder atm you cannot re-produce this issue anymore because of the commit above.  On dev https://dev.form-builder.lblod.info/formbuilder/65647FC498CA4C0008000036/edit/code you can still make a form with a listing and see that the inital create of the listing, the field will not have an order predicate (but I think you saw that :))

Rahien:
and i believe that with that change you made we can close this issue. Agreed @JonasVanHoof ?

PR #158, opened 2023-11-22
bump appuniversum to 2.18.0

nvdk:
this should resolve issues we've had with the embroider strict test scenario


Windvis:
The CI issue should be resolved by rebasing on #159 but the updated styles change might still be nice 👍.

I don't understand how ember-cli-sass can find the addon styles without configuring extra includePaths though 🤔. When I implemented the deprecation it looked like that was needed for importing styles from the node_modules folder. Maybe I misremember things.

PR #156, opened 2023-11-20
SFB-71: Update concept-scheme-selector

JonasVanHoof:
## ID
[ SFB-71](https://binnenland.atlassian.net/browse/SFB-71)

 ## Description

In the `form-builder` when selecting the `dropdown` component it wants to render it in our "preview" form. As the `concept-scheme-selector` is not provided with options yet it breaks the form and the application is unusable. 

These changes check if there are (valid) options proved to use in the component else it is just rendering a dropdown without options. 

 ## Type of change

 - [x] Bug fix  => for `form-builder`
 - [x] New feature => validation
 - [ ] Breaking change
 - [ ] Other

 ## How to test

In the dummy app remove or change the options of the dropdown and try to make the form break because of it.

Talking about this option selector..
```
##########################################################
# Dropdown
##########################################################
ext:conceptSchemeSelectorField a form:Field ;
    sh:name "Select" ;
    sh:order 270 ;
    sh:path ext:conceptSchemeSelectorValue ;
    form:options """{"conceptScheme":"http://example-concept-schemes/concept-schemes/foo-bar-baz","searchEnabled":false}""" ;
    form:displayType displayTypes:conceptSchemeSelector ;
    sh:group ext:mainPg .

ext:mainFg form:hasField ext:conceptSchemeSelectorField .
```

 ## Links to other PR's

 -  [Comment](https://github.com/lblod/frontend-form-builder/pull/55#pullrequestreview-1736708050) in form builder PR

PR #160, opened 2023-11-27
SFB-71: Show a warning when the field with options is missing option predicate

JonasVanHoof:
## ID
 [SFB-71](https://binnenland.atlassian.net/browse/SFB-71)

 ## Description

 When not passing any options to the fields that require them the form breaks. This means that if one component does not have the correct options passed to it breaks the whole form.

 ## Type of change

 - [x] Bug fix
 - [x] New feature
 - [x] Breaking change
 - [ ] Other

 ## How to test

Remove the `form:options` in the `.ttl` files for each field that has it as a predicate. this will show warnings on what is missing and not break the complete RDF-form.

 ## Links to other PR's

 - Same as https://github.com/lblod/ember-submission-form-fields/pull/156

### DIFF
--- a/addon/components/rdf-input-fields/concept-scheme-multi-select-checkboxes.js
+++ b/addon/components/rdf-input-fields/concept-scheme-multi-select-checkboxes.js
@@ -8,6 +8,7 @@ import {
   triplesForPath,
 } from '@lblod/submission-form-helpers';
 import { namedNode } from 'rdflib';
+import { hasValidFieldOptions } from '../../utils/has-valid-field-options';
 
 export default class RDFInputFieldsConceptSchemeMultiSelectCheckboxesComponent extends InputFieldComponent {
   @tracked options = [];
@@ -50,6 +51,11 @@ export default class RDFInputFieldsConceptSchemeMultiSelectCheckboxesComponent e
     const { sourceGraph, metaGraph } = this.args.graphs;
     const path = this.args.field.rdflibPath;
     const options = this.args.field.options;
+
+    if (!hasValidFieldOptions(this.args.field, ['conceptScheme', 'orderBy'])) {
+      return;
+    }
+
     const scheme = new namedNode(options.conceptScheme);
 
     let orderBy;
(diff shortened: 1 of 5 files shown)


addon/components/rdf-input-fields/concept-scheme-multi-select-checkboxes.js
```diff
@@ -50,6 +51,11 @@ export default class RDFInputFieldsConceptSchemeMultiSelectCheckboxesComponent e
     const { sourceGraph, metaGraph } = this.args.graphs;
     const path = this.args.field.rdflibPath;
     const options = this.args.field.options;
+
+    if (!hasValidFieldOptions(this.args.field, ['conceptScheme', 'orderBy'])) {
```
JonasVanHoof, line 55: I saw on two places that the `orderBy` is also present in the ttl files, ref: [subsidie-beheer](https://github.com/lblod/app-digitaal-loket/blob/fab8316a51e73e16f3af4a9218270dc3b63b74a6/config/subsidy-application-management/forms/e-inclusion/proposal/versions/20220804090000/form.ttl#L247) but this is not required for showing the component correctly.

Should we add it here as a required property?

addon/utils/has-valid-field-options.js
```diff
@@ -0,0 +1,30 @@
+export function hasValidFieldOptions(
+  fieldModel,
+  requiredProperties = ['conceptScheme']
```
JonasVanHoof, line 3: Should we remove this default? So it can also be used for ... if needed (not the case atm i think)
* `ext:alertField` => requires some properties (didn't check which)
* `ext:headerField` => not required to add them
* `ext:inputField` => not required to add them
* `ext:caseNumberField` => not required to add them (but will result in weird `null1234` output in dummy app)

JonasVanHoof: Actually should remove it as the name of the method does not state that this will check for the conceptScheme key as default. Leaving it atm till I have a response

addon/components/rdf-input-fields/concept-scheme-multi-select-checkboxes.js
Windvis: I wouldn't add it, no. The component seems to consider it optional.

addon/utils/has-valid-field-options.js
Windvis: Removing it makes sense, yea.

Windvis (approved):

PR #162, opened 2023-11-29
[LMB-34] feat: modelv2

Rahien:
transition to the final model v2:
- ~~move order to sections [todo]~~ this makes no sense, order is properly defined and works in the form builder as desired. I had misunderstood the extra indirection that was required by the form builder
- change propertygroups to sections [done]
- change sh:group to from:partOf [done]
- add new model example form [done]
- add rendersWhen constraint [done]

How to test:
- checkout this branch
- run the dummy application
- verify that the model for example form http://localhost:4200/dynamic-behaviour-v2 is indeed v2-target
- visit http://localhost:4200/dynamic-behaviour-v2 and verify that everything still works (new model)
- visit http://localhost:4200/dynamic-behaviour and verify that everything still works (old model)

image

Would need an approve of https://github.com/lblod/submission-form-helpers/pull/55 to also support the new way of expressing validations.


tests/dummy/public/test-forms/dynamic-behaviour-v2/form.ttl
```diff
@@ -0,0 +1,143 @@
...
+    form:includes fields:fd40cdcf-d085-4b71-930f-0b1c05e021d9 ;
+    form:includes fields:43dd5003-c9a4-48c6-b650-b90f2bce8c7d ;
+    form:includes fields:6fda9f15-6f5c-4174-bb28-3fac0804cefa ;
+    form:includes fields:star-rating-subset ;
+    form:includes fields:more-comment-subset .
+
+##########################################################
+#  property-group
+##########################################################
+fields:8e24d707-0e29-45b5-9bbf-a39e4fdb2c11 a form:Section;
```
JonasVanHoof, line 33: If possible I think we would like to get rid of the uuid's as a name and use naming in words instead for example this one could be 
```
fields:8e24d707-0e29-45b5-9bbf-a39e4fdb2c11 a form:Section;

# to

fields:singleCodeListValueS a form:Section;
```
Adding the First letter of the type at the end of the name is done in the formBuilder (in this example `S` stand for `Section`) not sure if we want to make it a general way of naming?

addon/helpers/is-section.js
```diff
@@ -0,0 +1,14 @@
...
+  PROPERTY_GROUP_DISPLAY_TYPE,
+  SECTION_DISPLAY_TYPE,
+} from '../models/section';
+
+export function isSection([displayType]) {
+  return (
+    displayType === SECTION_DISPLAY_TYPE ||
+    displayType === PROPERTY_GROUP_DISPLAY_TYPE
+  );
+}
```
JonasVanHoof, line 12: Like it ✅

addon/utils/model-factory.js
```diff
@@ -1,27 +1,27 @@
...
+    sectionHelpers.containingItem(field, { store, formGraph })
   );
 
-  const groups = mappedFields
+  const sections = mappedFields
     // .filter( (fieldGroup) => fieldGroup )
     .reduce((acc, item) => {
-      const pg = new PropertyGroup(item, { store, formGraph });
+      const pg = new Section(item, { store, formGraph });
       acc[item.value] = pg;
```
JonasVanHoof, line 25: Can this be done directly? Than we also get rid of the `pg` naming  :)
```
 acc[item.value] = new Section(item, { store, formGraph });;
 ```

File: addon/utils/model-factory.js [hunk 4]
```diff
@@ -33,16 +33,21 @@ function createPropertyTreeFromFields(
       sourceGraph,
       sourceNode,
     });
-    let groupUri = store.any(fieldUri, SHACL('group'), undefined, formGraph);
-    const group = groups[groupUri.value];
+    let sectionUri = sectionHelpers.containingItem(fieldUri, {
```
JonasVanHoof, line 36: nitpick: Not sure why but in my head `sectionHelpers` referes to a variable and `SectionHelpers` refers to a helper ? idk  😆 

(Maybe it comes from a previous project where helpers where classes and those where always starting with a capital letter)

addon/utils/model-factory.js
```diff
@@ -71,21 +87,19 @@ export function getTopLevelPropertyGroups({ store, graphs, form }) {
   if (
     getFormModelVersion(form, {
       store,
-      formGraph: graphs.formGraph,
+      formGraph: formGraph,
     }).toLowerCase() == 'v2'
```
JonasVanHoof, line 91: nitpick: Should we make a constantout of the version or is this a one time thing to use here?
Thinking `v2` or `V2` or `version2`

addon/utils/model-factory.js
```diff
@@ -151,45 +152,55 @@ export function getSubFormsForNode(
...
     metaGraph: graphs.metaGraph,
     sourceNode: node,
   });
 
-  // NOTE: contains all children for a property-group (this can include other nested property-groups)
-  const children = store
-    .match(undefined, SHACL('group'), group.uri, graphs.formGraph)
+  // NOTE: contains all children for a section (this can include other nested sections)
+  const children = sectionHelpers
+    .getItems(section.uri, { store, formGraph })
```
JonasVanHoof, line 183: nitpick: you refer to `getItems` here but they will be stored in the `const children` should we rename the helper method to `getChildren()`?

JonasVanHoof (commented):
Just some thoughts and nitpicking good job already! 💪🏻

tests/dummy/public/test-forms/dynamic-behaviour-v2/form.ttl
Rahien: added a sections: prefix for extra clarity

tests/dummy/public/test-forms/dynamic-behaviour-v2/form.ttl
```diff
@@ -0,0 +1,143 @@
...
+        form:customValue <http://redpencil.data.gift/concepts/0d5b3b23-c055-4d56-83d0-b65d28050315>
+      ] .
+
+
+##########################################################
+#  property-group: subset of star rating
+##########################################################
+fields:subset-property-group a form:Section;
+    sh:description "parent property-group, used to group fields and property-groups together";
+    sh:name "Select a value, if it matches subset of ratings it will trigger a text area" ;
```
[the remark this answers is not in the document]
Rahien: nice find!

tests/dummy/public/test-forms/dynamic-behaviour-v2/form.ttl
```diff
@@ -0,0 +1,143 @@
...
+    ##########################################################
+    form:rendersWhen
+      [ a form:MatchValues ;
+        form:grouping form:Bag ;
+        sh:path ext:star-rating-subset-property ;
+        # Tip: you can use form:valuesNotIn to exclude from a list (if triples are found matching the path)
+        form:valuesIn (
+          <http://redpencil.data.gift/concepts/df2b8b6e-bcc2-4608-83da-98ab251fcf54>
+          <http://redpencil.data.gift/concepts/d3bcb263-c4e4-4512-98c8-d0b45400eebd> );
+      ] .
```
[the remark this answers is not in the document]
Rahien: can't take credit for it though, it was already present in the dynamic-behaviour form.ttl \o/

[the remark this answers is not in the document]
JonasVanHoof: Hehe the code yes but the `form:rendersWhen` L134 didn't I think? Overall the naming of the predicates feels more connected to each other.

tests/dummy/public/test-forms/dynamic-behaviour-v2/form.ttl
```diff
@@ -0,0 +1,144 @@
...
+@prefix schema: <http://schema.org/>.
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>.
+@prefix sections: <http://data.lblod.info/fields/sections/> .
+
+##########################################################
+# form
+##########################################################
+
+form:main a form:Form, form:TopLevelForm ;
+    mu:uuid "6b70a6f0-cce2-4afe-81f5-5911f45b0b27" ;
```
JonasVanHoof, line 21: remove uuid since we use the named nodes

tests/dummy/public/test-forms/dynamic-behaviour-v2/form.ttl
```diff
@@ -0,0 +1,144 @@
...
+    form:includes fields:comment-one-star ;
+    form:includes fields:comment-two-stars ;
+    form:includes fields:star-rating-subset ;
+    form:includes fields:more-comment-subset .
+
+##########################################################
+#  property-group
+##########################################################
+sections:main a form:Section;
+    mu:uuid "8e24d707-0e29-45b5-9bbf-a39e4fdb2c11";
```
JonasVanHoof, line 35: remove uuid

tests/dummy/public/test-forms/dynamic-behaviour-v2/form.ttl
```diff
@@ -0,0 +1,144 @@
...
+    sh:description "parent property-group, used to group fields and property-groups together";
+    sh:name "Single code list value" ;
+    sh:order 1 .
+
+##########################################################
+# Dropdown
+##########################################################
+fields:main-star a form:Field ;
+    mu:uuid "fd40cdcf-d085-4b71-930f-0b1c05e021d9";
```
JonasVanHoof, line 44: remove uuid

tests/dummy/public/test-forms/dynamic-behaviour-v2/form.ttl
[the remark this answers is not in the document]
JonasVanHoof: Works 🚀

JonasVanHoof (approved):
Everything seems to work correctly with the new and old version. Looks good, thanks! 🚀

tests/dummy/public/test-forms/dynamic-behaviour-v2/form.ttl
```diff
@@ -0,0 +1,144 @@
...
+    form:partOf fields:subset-property-group .
+
+
+##########################################################
+# More comment subset stars rating
+##########################################################
+fields:more-comment-subset a form:Field ;
+    sh:name "More comments about star ratings 3,4" ;
+    sh:order 220 ;
+    sh:path ext:ratingComment ;
```
Windvis, line 129: I would use a different path for this one. At the moment it overrides the value of the first comment fields which is a bit confusing. The issue is already there in the original form, but I noticed it while testing this PR.

tests/dummy/app/templates/application.hbs
```diff
@@ -18,11 +18,21 @@
                   Dynamic behaviour
                 </AuNavigationLink>
               </li>
+              <li class="au-c-list-navigation__item">
+                <AuNavigationLink @route="form" @model="dynamic-behaviour-v2">
+                  Dynamic behaviour v2
```
Windvis, line 23: Nitpick, but maybe we should rename the old version to `Dynamic behavior (legacy)` with a disclaimer that we shouldn't be using that setup anymore?

This one could then simply be `Dynamic behavior` or `Conditional rendering` or something?

tests/dummy/app/templates/application.hbs
```diff
@@ -57,6 +67,11 @@
                   Tables
                 </AuNavigationLink>
               </li>
+              <li class="au-c-list-navigation__item">
+                <AuNavigationLink @route="form" @model="validations-v2">
+                  Validations V2
```
Windvis, line 72: Same here I guess

addon/components/section.hbs
Windvis, line 1: Hmm, the issue with the current setup is that it exposes components in the "global" app scope due to the way Ember worked in the past. Ember is slowly moving away from that with strict templates but we haven't updated the addon code yet.

"Section" is the correct name in the context of this addon, but I fear it might be too generic and some apps might already have a `Section` component (or they might want to create one in the future).

Some options:
1. Double check that we don't have a `Section` component in any of the apps that are using this addon
2. Rename the component to FormSection which should be specific enough
3. Don't pollute the global scope and use references instead by importing the component and storing it on a class property so it can be used as reference in the template. This makes it truly "private" to the addon and is in line with the future of Ember. 

I obviously prefer option 3 but it's the most work (and probably requires some experimentation with the recursive component invocations).

Windvis: Option 2 would be the quickest while still being somewhat safe, but maybe I'm just being overly paranoid and `Section` is good enough until we switch to the strict mode version.

Thoughts?

tests/dummy/public/test-forms/validations-v2/form.ttl
Windvis, line 1: This seems to be based on the "basic fields" form. I would just update the validation setup in that one. I don't think we want to promote the v1 version once v2 is supported?

addon/helpers/is-section.js
Windvis, line 1: Similar to components, helpers are also globally accessible by default. Since this is a private helper that is only used by the Section component, I would move it into the component .js file and store it on a class property. It can then be used as a reference in the template, and we don't pollute the global helper scope anymore.

```js
//section.js
const isSection = helper(function([displayType]) {...})

class SectionComponent extends ... {
  isSection = isSection;
}
```

```hbs
{{! section.hbs}}
{{#if (this.isSection field.displayType)}}
```

addon/utils/display-type.js
```diff
@@ -206,6 +207,10 @@ registerComponentsForDisplayType([
     displayType: 'http://lblod.data.gift/display-types/property-group',
     edit: PropertyGroupComponent,
   },
+  {
+    displayType: 'http://lblod.data.gift/display-types/section',
+    edit: SectionComponent,
+  },
```
Windvis, line 213: Hmmmm, I think this isn't needed and the property-group version is probably a mistake I made in the past. Do we use these display types anywhere? I think we retrieve all the sections with a helper in the relevant components?

Windvis (commented):
Thanks for working on this!

tests/dummy/public/test-forms/dynamic-behaviour-v2/form.ttl
Rahien: not sure what you mean here, i just copied this from the dynamic-behaviour example. I agree that it's not really necessary that these predicates are there but i thought 'it can't hurt' would you like me to remove them in the old example too?

addon/helpers/is-property-group.js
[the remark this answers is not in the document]
Rahien: nice! removing!

addon/utils/model-factory.js
```diff
@@ -71,21 +84,19 @@ export function getTopLevelPropertyGroups({ store, graphs, form }) {
   if (
     getFormModelVersion(form, {
       store,
-      formGraph: graphs.formGraph,
-    }).toLowerCase() == 'v2'
+      formGraph: formGraph,
+    }).toLowerCase() == V2
```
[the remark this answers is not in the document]
Rahien: changed it mostly as a reflex because coming from ts, you'd get some automatic shouting if you tried to use an import that doesn't exist... no such luck in javascript of course...  Still feels a bit magic constant to repeat a string throughout the code no?

Not sure about the history of this file either but a refactor a big as this would probably warrant it's own PR?

addon/utils/model-transition.js
```diff
@@ -0,0 +1,43 @@
+import { RDF, FORM, SHACL } from '@lblod/submission-form-helpers';
+/**
+ * sh:group was replaced by form:partOf and form:PropertyGroup was replaced by form:Section
+ */
+export const SectionHelpers = {
```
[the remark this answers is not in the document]
Rahien: will change!

addon/utils/model-factory.js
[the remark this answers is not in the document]
Windvis: > Still feels a bit magic constant to repeat a string throughout the code no?

For sure, I just don't think that moving something into a constant was the "fix" for this magic string here. It _feels_ worse to use a constant and then having to know what it contains instead of just using the magic string. If for some reason the constant would change, this code would need to be updated as well. I probably would have made a function that did the check instead, but that seems to exist in the helper package already (but not public).

In any case, it's not worth changing it back either, so just ignore this remark. Sorry for the noise.

> Not sure about the history of this file either but a refactor a big as this would probably warrant it's own PR?

Yea, I was just thinking out loud about what the "next steps" could be.

addon/components/rdf-input-fields/input-field.js
```diff
@@ -44,9 +44,23 @@ export default class InputFieldComponent extends Component {
...
+      // fall back to old model:
+      constraints = store.match(
+        this.args.field.uri,
+        FORM('validations'),
+        undefined,
+        formGraph
+      );
+    }
+
+    return constraints.map((t) => t.object);
```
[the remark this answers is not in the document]
Rahien: indeed. it seems that the validationcontraints getter is only used in the maxLength getter which checks if there is a max length contrains. However, it then also needs the contraint's uri because it needs  the max value for the contraint... So if we want to refactor this, we need to change the helpers indeed. I will add a pr there soon to add a method returning a mapping of validation constraint uri -> validation type uri so you have both if you want

addon/components/section.js
```diff
@@ -0,0 +1,141 @@
...
+        form: this.args.form,
+        store: this.args.formStore,
+        graphs: this.args.graphs,
+        node: this.args.sourceNode,
+      });
+    });
+  }
+
+  get section() {
+    return this.args.section || this.args.group;
```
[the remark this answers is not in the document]
Rahien: it's done in case of someone using our component (or the propertygroup mirror component) in the 'old' way, passing in @group

tests/dummy/public/test-forms/dynamic-behaviour-v2/form.ttl
Rahien: had the same confusing experience while writing this but wanted to be as close as possible to the old example. Now that i see that I'm not alone in finding this annoying i will just change it in both examples

addon/components/section.hbs
Rahien: Will go with option 2, option 3 sounds like what we should do at some point in the future but isn't the point of this PR

addon/helpers/is-section.js
Rahien: alright! i removed both this helper and the is-listing helper as well.

addon/utils/display-type.js
Rahien: seems you are right, the section component is handled explicitly where the component type for other components is determined dynamically if it's not a section or listing

addon/components/rdf-input-fields/input-field.js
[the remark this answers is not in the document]
Rahien: The PR that will allow this simplification: https://github.com/lblod/submission-form-helpers/pull/61

addon/components/section.js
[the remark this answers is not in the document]
Windvis: I would still switch back to using the argument directly in the template, or remove the this.args.group part since it's not needed.

addon/components/form-section.js
```diff
@@ -0,0 +1,159 @@
...
+      child &&
+      [SECTION_DISPLAY_TYPE, PROPERTY_GROUP_DISPLAY_TYPE].includes(
+        child.displayType
+      )
+    );
+  }
+
+  childIsListing(child) {
+    return child && child.displayType === LISTING_TYPE;
+  }
```
Windvis, line 90: We still officially support Ember 3.28 in this addon, so this won't work unless we also install the https://github.com/ember-polyfills/ember-functions-as-helper-polyfill as a dependency.

I would just use the helper wrapper instead for now.

tests/dummy/app/templates/application.hbs
```diff
@@ -57,6 +70,11 @@
...
+                </AuNavigationLink>
+              </li>
```
Windvis, line 77: This link is no longer needed now.

Windvis (approved):
Awesome, thanks!

PR #164, opened 2023-11-30
bump helpers to allow negative constraints

Rahien:
This PR bumps the helpers package to allow the use of negative path constraints
It also adds example form which is a modified form builder that shows table and regular listings are no longer confused regarding their scope (which previously resulted in duplicate sections being rendered in the incorrect listing).


Rahien:
closing and including in https://github.com/lblod/ember-submission-form-fields/pull/162

PR #166, opened 2023-12-13
Provided the an extra case in dummy forms that need support in v2 model

cecemel:
To be fully v2 model compliant we need to make sure the form:validatedBy is supported.
Still hold your horses to resolve discussion about the cost of renaming things.


Windvis:
I don't think this one is still needed now that #162 is merged. Closing, feel free to reopen if the assumption was wrong.

PR #169, opened 2024-01-15
Add support for Appuniversum v3

Windvis:


package.json
```diff
@@ -92,6 +92,7 @@
     "stylelint": "^15.4.0",
     "stylelint-config-standard": "^32.0.0",
     "stylelint-prettier": "^3.0.0",
+    "tracked-built-ins": "^3.3.0",
```
Windvis, line 95: This is a peerdep for ember-file-uploads, without this the app doesn't work. We don't need to list it as a peerDep since we already list Appuniversum.

PR #171, opened 2024-01-18
Add basic support for "warning" validations

Windvis:
Depends on https://github.com/lblod/submission-form-helpers/pull/62.


Windvis:
Once this is merged we should add a section to the wiki about validation "severity".

PR #173, opened 2024-01-30
Switch to Woodpecker CI

Windvis:
This replaces Drone with Woodpecker, based on the [template repo](https://github.com/redpencilio/ci-templates/tree/master/woodpecker/ember).


.woodpecker/release.yml
Windvis, line 1: Whoops, this is wrong. It needs to be a pipeline that publishes to npm..

PR #175, opened 2024-02-07
Hide labels in "show" mode

Windvis:
This hides all the "required" and "remaining character" labels when the form is in the read-only state. It fixes a regression that we introduced when we ["deduped" the fields](https://github.com/lblod/ember-submission-form-fields/pull/54).

**To test:**
- Flip the "read-only mode" toggle in the top bar
- verify that there are no "required" or "remaining character" labels (you can temporarily add extra constraints to some fields if you want)


JonasVanHoof (approved):
Goes from 
image

To
image

Looks good

Code wise I don't think i could've done it in any other way :D

PR #177, opened 2024-02-07
SFB-125: Euro masking input

JonasVanHoof:
Use the input mask library directly instead of the input mask modifier


JonasVanHoof:
Final implementation is merged here https://github.com/lblod/ember-submission-form-fields/commit/e0ffb4d7b366a39b7d9f944cc04b0494eabfa22b

PR #181, opened 2024-02-29
Add bulk download support

Windvis:
This adds bulk download support to the `RemoteUrl` and `Files` fields. The feature is only enabled in "read only" mode.

### Test app
The test app includes some mock data so it can be used to quickly test it.

> individual file downloads don't work as expected since those can't be mocked by msw

### Loket / other toezicht based apps
- Link this PR into the frontend of the app you want to test with yalc or npm link (untested)
- Proxy to an environment that has some submitted submission with more than 1 (downloadable) link or file
- Press the "download as zip" button
- verify that all the files are in the zip


Riadabd (approved):
Looks good! The downloads work perfectly on the test app, and the download button is only visible in read-only mode, as expected :+1:

addon/components/rdf-input-fields/files.js
```diff
@@ -166,4 +177,21 @@ export default class RdfInputFieldsFilesComponent extends InputFieldComponent {
     this.files.removeObject(fileField);
     // general validation of the field is handled by onStoreUpdate()
   }
+
+  downloadAsZip = dropTask(async () => {
+    const promises = this.files.map((file) => {
+      return fetch(file.record.downloadLink);
```
Windvis, line 183: One thing I forgot, and just noticed.. `fetch` doesn't consider 40* responsen to be errors, so 404s don't make the promise reject either 😄.

So the zip will contain a file with the correct name, but its contents will be the '404, not found' text.

PR #182, opened 2024-03-05
SFB-206: POC show fields without link to group

JonasVanHoof:
## ID
 [SFB-206](https://binnenland.atlassian.net/browse/SFB-206)

 ## Description

We want to show fields with links in the form. When adding a field in the builder you must assign the link to property for that field to be able to show the field in the preview. 

This change makes it that the "toplevel" fields will be shown even when they are not linked to a section/propertygroup

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [x] Breaking change
 - [ ] Other

 ## How to test

1. Go to the builder and add a field
2. Do not assign the link to property in the builder
3. The field should show in the form preview

 ## Links to other PR's

 - /

 ## Attachments

**v1** 

image

### DIFF
--- a/addon/components/rdf-form.hbs
+++ b/addon/components/rdf-form.hbs
@@ -20,6 +20,30 @@
           {{!Temporary argument so we don't break the css hacks in the LPDC module}}
         />
       {{/if}}
+      {{#if (this.isFieldModel model)}}
+        <div class="au-u-margin-left">
+          {{#let
+            (component-for-display-type model.displayType show=@show)
+            as |FormField|
+          }}
+            <FormField
+              @field={{model}}
+              @form={{@form}}
+              @formStore={{@formStore}}
+              @graphs={{@graphs}}
+              @sourceNode={{@sourceNode}}
+              @forceShowErrors={{@forceShowErrors}}
+              @cacheConditionals={{@cacheConditionals}}
+              @show={{@show}}
+            />
+          {{/let}}
+
+          {{#if model.help}}
+            {{! template-lint-disable no-triple-curlies}}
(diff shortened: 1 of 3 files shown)


Windvis:
@aatauil @JonasVanHoof I can review the code, but the underlying change itself is a modification/extension of the data model, no? Was this discussed somewhere? I can't evaluate if this is something we want here myself. Looking at the description it _feels_ like something that is only needed because of the meta-form setup in the form-builder? (Not saying this couldn't simplify other forms, but there was no _need_ for it until now, AFAIK).

addon/components/rdf-form.hbs
```diff
@@ -1,19 +1,51 @@
...
+      {{/if}}
+      {{#if (this.isFieldModel model)}}
+        <div
+          class="au-o-flow au-u-margin-left au-u-margin-right au-u-margin-top"
+        >
+          {{#let
+            (component-for-display-type model.displayType show=@show)
+            as |FormField|
+          }}
+            <FormField
```
Windvis, line 31: I haven't looked at the implementation of the validation helper, but do these fields still validate correctly? If not, we probably need the same lookup logic in the helpers package?

JonasVanHoof: Good question! Didn't think of adding a field with validations to the toplevel form as in the formbuilder you can only add validations to a field if they are in a section (if you click a form togheter)

The formField code was taken from the form-section code so I made an assumption that it would do everyting the same as in that component.

Will be checking this when picking up the ticket again 🙏🏻

JonasVanHoof:
We will not be showing the "toplevel" fields in the RDF form. For the formbuilder we will just assign the newly added field to the first section in the form.

PR #183, opened 2024-03-12
Handle "not ok" responses while preparing the bulk zip files

Windvis:
Small follow-up to #181 

We forgot to handle the cases where `fetch` doesnt't throw, but the server doesn't send an "ok" response either. The most probably case will be 404 errors. Now we display the error message if the file doesn't exist.

(This doesn't need to appear in the changelog since the previous PR hasn't been released yet)

**To test**
- run the test app
- rename the dummy files in `tests/dummy/public/mock-files` to something else
- try to download the zips by clicking the buttons in the test app
- it should display an error message (in a toast message and the console)

### DIFF
--- a/addon/components/rdf-input-fields/files.js
+++ b/addon/components/rdf-input-fields/files.js
@@ -36,6 +36,7 @@ class FileField {
 
 export default class RdfInputFieldsFilesComponent extends InputFieldComponent {
   @service store;
+  @service toaster;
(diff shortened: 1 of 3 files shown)


Windvis:
Merged it already since I'm pretty confident that it works, if there are any remarks about the implementation I'll resolve those in a follow-up PR.

PR #184, opened 2024-03-26
SFB-248: radio button concept scheme selector with order by property in form:options

JonasVanHoof:
## Issue
 [SFB-248](https://binnenland.atlassian.net/browse/SFB-248)

## Description

When the `form:options` have `orderBy` property defined sort the concepts by that order instead of `byLabel` (the default). 

## Testing

Go to the basic fields tab and look at the radio buttons with orderBy those are sorted by the orderBy predicate.

## Note

What bothers me is that we assume that `orderBy` is a number. @Rahien  should we do it now like this so we can have a look at a order predicate for sorting these in the future?

### DIFF
--- a/addon/components/rdf-input-fields/concept-scheme-multi-select-checkboxes.js
+++ b/addon/components/rdf-input-fields/concept-scheme-multi-select-checkboxes.js
@@ -47,50 +47,66 @@ export default class RDFInputFieldsConceptSchemeMultiSelectCheckboxesComponent e
   }
 
   loadOptions() {
-    const store = this.args.formStore;
-    const { sourceGraph, metaGraph } = this.args.graphs;
     const path = this.args.field.rdflibPath;
-    const options = this.args.field.options;
+    const fieldOptions = this.args.field.options;
+    let orderBy = null;
 
     if (!hasValidFieldOptions(this.args.field, ['conceptScheme'])) {
       return;
     }
 
-    const scheme = new namedNode(options.conceptScheme);
+    if (hasValidFieldOptions(this.args.field, ['orderBy'])) {
+      orderBy = new namedNode(fieldOptions.orderBy);
+    }
 
-    let orderBy;
-    if (options.orderBy) orderBy = new namedNode(options.orderBy);
-
-    this.options = store
-      .match(undefined, SKOS('inScheme'), scheme, metaGraph)
+    const conceptScheme = new namedNode(fieldOptions.conceptScheme);
+    this.options = this.store
+      .match(undefined, SKOS('inScheme'), conceptScheme, this.graphs.metaGraph)
       .map((t) => {
         const subject = t.subject;
-        const label = store.any(
+        const label = this.store.any(
           subject,
           SKOS('prefLabel'),
           undefined,
-          metaGraph
+          this.graphs.metaGraph
         ).value;
-        const provided = !!store.any(
+        const provided = !!this.store.any(
           this.storeOptions.sourceNode,
           path,
           subject,
-          sourceGraph
+          this.graphs.sourceGraph
         );
-
-        let order = 0;
-        if (orderBy)
-          order = parseInt(
-            store.any(subject, orderBy, undefined, metaGraph).value
-          );
 
         return {
           subject,
           label,
           provided,
-          order,
+          order: this.getOrderForOption(orderBy, t.subject),
         };
       });
-    this.options.sort((a, b) => a.order - b.order);
+
+    this.options.sort((a, b) =>
+      a.order.localeCompare(b.order, undefined, { numeric: true })
+    );
+  }
+
+  getOrderForOption(orderBy, tripleSubject) {
+    const orderStatement = this.store.any(
+      tripleSubject,
+      orderBy,
+      undefined,
+      this.graphs.metaGraph
+    );
+
+    // must be string because above we are using string.localCompare
+    return `${orderStatement?.value ?? ''}`;
+  }
+
+  get store() {
+    return this.args.formStore;
+  }
+
+  get graphs() {
+    return this.args.graphs;
   }
 }
--- a/addon/components/rdf-input-fields/concept-scheme-radio-buttons.js
+++ b/addon/components/rdf-input-fields/concept-scheme-radio-buttons.js
@@ -64,6 +64,7 @@ export default class RdfInputFieldsConceptSchemeRadioButtonsComponent extends Si
     );
 
     if (orderStatement) {
+      // This is also passing NaN if the value is a string
       return parseInt(orderStatement.value);
     }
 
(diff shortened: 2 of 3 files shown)


addon/components/rdf-input-fields/concept-scheme-radio-buttons.js
```diff
@@ -14,40 +14,74 @@ export default class RdfInputFieldsConceptSchemeRadioButtonsComponent extends Si
...
+  getOrderForOption(orderBy, tripleSubject) {
+    const orderStatement = this.args.formStore.any(
+      tripleSubject,
+      orderBy,
+      undefined,
+      this.metaGraph
+    );
+
+    if (orderStatement) {
+      // This is also passing NaN if the value is a string
```
Rahien, line 67: could we simply check if this is NaN and fall back to the string form if it is?

addon/components/rdf-input-fields/concept-scheme-radio-buttons.js
```diff
@@ -14,40 +14,74 @@ export default class RdfInputFieldsConceptSchemeRadioButtonsComponent extends Si
...
+      return parseInt(orderStatement.value);
```
Rahien, line 68: i would not do this and use the change in byOrder below, wdyt?

addon/components/rdf-input-fields/concept-scheme-radio-buttons.js
```diff
@@ -14,40 +14,74 @@ export default class RdfInputFieldsConceptSchemeRadioButtonsComponent extends Si
...
 }
 
 function byLabel(a, b) {
   const textA = a.label.toUpperCase();
   const textB = b.label.toUpperCase();
   return textA < textB ? -1 : textA > textB ? 1 : 0;
 }
+
+function byOrder(a, b) {
+  return a.order - b.order;
```
Rahien, line 86: I would do
```suggestion
  return a.order.localeCompare(b.order, undefind, {numeric: true});
```
with checking on a.order being undefined/null
That way, we don't care what the value of the property is and we support sorting by some label too if we want

Rahien (changes requested):

addon/components/rdf-input-fields/concept-scheme-radio-buttons.js
JonasVanHoof: This is also a relieve if people pass a predicate that for example is the sh:name that the ordering will not break anything 🚀

addon/components/rdf-input-fields/concept-scheme-radio-buttons.js
JonasVanHoof: Used this suggestion

tests/dummy/public/test-forms/basic-fields/form.ttl
```diff
@@ -119,6 +119,19 @@ ext:radioButtonsField a form:Field ;
...
+ext:radioButtonsWithOrderByField a form:Field ;
+    sh:name "Radio buttons with orderBy" ;
+    sh:order 211 ;
+    sh:path ext:radioButtonsValue ;
+    form:options """{"conceptScheme":"http://example-concept-schemes/concept-schemes/foo-bar-baz", "orderBy": "http://purl.org/linked-data/cube#order"}""" ;
+    form:displayType displayTypes:conceptSchemeRadioButtons ;
+    sh:group ext:mainPg .
+
+ext:mainFg form:hasField ext:radioButtonsWithOrderByField .
+
```
JonasVanHoof, line 134: @Rahien  should I leave this in?

addon/components/rdf-input-fields/concept-scheme-radio-buttons.js
```diff
@@ -14,40 +14,73 @@ export default class RdfInputFieldsConceptSchemeRadioButtonsComponent extends Si
...
+  getOrderForOption(orderBy, tripleSubject) {
+    const orderStatement = this.args.formStore.any(
+      tripleSubject,
+      orderBy,
+      undefined,
+      this.metaGraph
+    );
+
+    if (orderStatement) {
+      return orderStatement.value;
```
Rahien, line 67: would play it safe and wrap in a string else the localeCompare won't work if it's a number

tests/dummy/public/test-forms/basic-fields/form.ttl
Rahien: I would. It's nice to be able to find an example of things in this dummy app (helped me out quite a few times now)

Rahien (changes requested):

addon/components/rdf-input-fields/concept-scheme-multi-select-checkboxes.js
```diff
@@ -47,50 +47,66 @@ export default class RDFInputFieldsConceptSchemeMultiSelectCheckboxesComponent e
...
-          order,
+          order: this.getOrderForOption(orderBy, t.subject),
         };
       });
-    this.options.sort((a, b) => a.order - b.order);
+
+    this.options.sort((a, b) =>
+      a.order.localeCompare(b.order, undefined, { numeric: true })
+    );
+  }
```
JonasVanHoof, line 91: adding the sort directly here as there is only one sorting

addon/components/rdf-input-fields/concept-scheme-multi-select-checkboxes.js
```diff
@@ -47,50 +47,66 @@ export default class RDFInputFieldsConceptSchemeMultiSelectCheckboxesComponent e
   }
 
   loadOptions() {
-    const store = this.args.formStore;
-    const { sourceGraph, metaGraph } = this.args.graphs;
     const path = this.args.field.rdflibPath;
```
JonasVanHoof, line 50: apparently this is not working when we use the getter method of the field model `.path`.

Will not look into this now

addon/components/rdf-input-fields/concept-scheme-radio-buttons.js
```diff
@@ -14,40 +14,74 @@ export default class RdfInputFieldsConceptSchemeRadioButtonsComponent extends Si
...
+
+  getOrderForOption(orderBy, tripleSubject) {
+    const orderStatement = this.store.any(
+      tripleSubject,
+      orderBy,
+      undefined,
+      this.metaGraph
+    );
+
+    // This MUST be a string so our byOrder sorting function returns the correct result
```
JonasVanHoof, line 66: Added the comment :)

addon/components/rdf-input-fields/concept-scheme-radio-buttons.js
```diff
@@ -14,40 +14,74 @@ export default class RdfInputFieldsConceptSchemeRadioButtonsComponent extends Si
...
+    return `${orderStatement?.value ?? ''}`;
+  }
+
+  get metaGraph() {
+    return this.args.graphs.metaGraph;
+  }
+
+  get store() {
+    return this.args.formStore;
+  }
```
JonasVanHoof, line 76: As the store is used in the load options method but also in another add a getter for it

JonasVanHoof (commented):
@Rahien  updated the checkboxes component and for the radio buttons added the comment + new getter for the store

Rahien (approved):

PR #185, opened 2024-04-04
SFB-224: Required validation on currency input (fix)

JonasVanHoof:
## Issue
 [SFB-224](https://binnenland.atlassian.net/browse/SFB-224)

## Description

The required validation on the currency input was not triggered when the user cleared the form as the value 0 was still in the field.

Now the value is set to null so the validation is triggered. When a user filled in the value **0** the required validation was also triggered as the value of the triple was not updated to zero but to no value. I found out that this was working with the numerical input and that they fixed it by updating the value as a literal with the integer or decimal datatype. 

The field uses the `hasBeenFocused` & `validation` check from the `SimpleInputField` but the currency input will specifically update the value to NULL when nothing is filled in.

## Testing

The currencyInput is added to the validation section here 
image

 1. Validation should not be visible on initial load
 2. When focused and no value filled in the error should be triggered
 3. Its possible to fill in the value `0` and not trigger the validation
 4. When a correct value is filled in no error is triggered

### DIFF
--- a/addon/components/rdf-input-fields/currency-input.js
+++ b/addon/components/rdf-input-fields/currency-input.js
@@ -1,6 +1,8 @@
 import { action } from '@ember/object';
 import { guidFor } from '@ember/object/internals';
 import SimpleInputFieldComponent from '@lblod/ember-submission-form-fields/components/rdf-input-fields/simple-value-input-field';
+import { XSD } from '@lblod/submission-form-helpers';
+import { literal } from 'rdflib';
 
 export default class RdfInputFieldsCurrencyInputComponent extends SimpleInputFieldComponent {
   inputId = 'input-' + guidFor(this);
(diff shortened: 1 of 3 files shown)


addon/components/rdf-input-fields/currency-input.js
```diff
@@ -9,12 +11,30 @@ export default class RdfInputFieldsCurrencyInputComponent extends SimpleInputFie
 
   @action
   updateValue(e) {
+    let updatedValue = null;
     if (e && typeof e.preventDefault === 'function') e.preventDefault();
 
     if (e.target && e.target.value) {
       this.value = e.target.inputmask.unmaskedvalue();
+      updatedValue = literal(Number(this.value), this.datatype);
```
JonasVanHoof, line 19: Should we just return the value as an integer always? `inputmask.unmaskedValue` should always return a number here ( https://robinherbots.github.io/Inputmask/#/documentation/numeric  and  ctrl+f  "unmaskAsNumber"). Than we can remove the `get datatype`

Windvis: Always a decimal you mean?

JonasVanHoof: Yes a decimal indeed
image

But with that in mind we cannot do do it because a decimal is not an integer 😄

PR #186, opened 2024-04-09
Ben/lmb 184 numerical input with clear

bfidlers:
Screenshot 2024-04-09 at 11 26 25


addon/styles/addon.css
```diff
@@ -96,3 +96,30 @@
...
+  cursor: pointer;
+  width: 2rem;
+  height: 2rem;
+  color: var(--au-gray-700);
+  background-color: var(--au-gray-200);
+  border-radius: 100%;
+  margin-left: -2.5rem;
+}
+
+.number-input-no-spin::-webkit-inner-spin-button,
```
Rahien, line 115: i don't know if we want to remove this. E.g. on mobile it could be very nice to have in some cases, no?

bfidlers: I only removed it if you have the option allowClear set to true, so in other cases it is still there. Seemed better UI wise, but I can certainly revert this commit.

addon/components/rdf-input-fields/numerical-input.hbs
```diff
@@ -8,17 +8,30 @@
...
+    @width={{if @inTable "block"}}
+    @warning={{this.hasWarnings}}
+    id={{this.inputId}}
+    lang="nl-BE"
+    type="number"
+    value={{this.value}}
+    {{on "change" this.updateValue}}
+  />
+  {{#if @field.options.allowClear}}
+    <AuIcon
```
Windvis, line 26: Is this button needed? Wouldn't clearing the field like any other input field be more intuitive? I would also find it strange to only see this clear button for numerical inputs, but non of the other fields.

addon/styles/addon.css
```diff
@@ -96,3 +96,30 @@
 .sf-files-card--small > div.au-u-margin-top-tiny {
   margin-top: 0 !important;
 }
+
+.center-container {
```
Windvis, line 100: Appuniversum has some [flex utils](https://appuniversum.github.io/ember-appuniversum/?path=/story/utilities-flex--page), so you could use those if you want.

addon/components/rdf-input-fields/numerical-input.hbs
JonasVanHoof: Should we just add a quick return here https://github.com/lblod/ember-submission-form-fields/blob/master/addon/components/rdf-input-fields/simple-value-input-field.js#L50 I would look something like this

```js
get CanClearInputField() {
   return ... <LOGIC>
}

updateValue(value) {
    if(!value && this.CanClearInputField) {
        updateSimpleFormValue(this.storeOptions, null, this.nodeValue);
        super.updateValidations();
        
        return;
     }
}
```


All inputs that are extending from `SimpleValueInputFieldComponent` will have the option to clear.
Just a thought :)

bfidlers: This seemed like the easier approach, but it might indeed be a bit confusing, will try your aproach.

addon/components/rdf-input-fields/numerical-input.js
```diff
@@ -11,8 +11,13 @@ export default class RdfInputFieldsNumericalInputComponent extends SimpleInputFi
   updateValue(e) {
     if (e && typeof e.preventDefault === 'function') e.preventDefault();
     this.value = e.target.value;
-    const number = literal(Number(this.value), this.datatype);
-    super.updateValue(number);
+    if (this.value == '' && this.args.field.options.allowClear) {
```
Windvis, line 14: Does this actually need to be an option, this sounds more like a bugfix to me, actually? Is there ever a use-case where you don't want this to happen?

bfidlers: We wanted to do this because of backwards compatibility, since this has been in the codebase for so long, it could be specific applications expect this behaviour.

addon/components/rdf-input-fields/numerical-input.js
```diff
@@ -1,18 +1,51 @@
...
+  loadOptions() {
+    let allowClear = this.args.formStore.any(
+      this.args.field.uri,
+      FORM_OPTION('allowClear'),
+      undefined,
+      this.args.graphs.formGraph
+    );
+    if (!allowClear) {
+      if (!hasValidFieldOptions(this.args.field, ['allowClear'])) {
+        return;
```
JonasVanHoof, line 30: nitpick:
Do we want to set this.allowClear  equal to false here? I always find it difficult when a value is defined on init.
```js
@tracked allowClear:
loadOptions() { 
  if (!hasValidFieldOptions(this.args.field, ['allowClear'])) {
        this.allowClear = false;
  }
}

JonasVanHoof (commented):
Only went through the code

addon/components/rdf-input-fields/numerical-input.js
bfidlers: fair point

addon/components/rdf-input-fields/numerical-input.js
cecemel: I follow @Windvis here. I wouldn't add this extra option. It was basically a bug at the start. Maybe put it in a release note?

PR #188, opened 2024-04-12
LMB-326: Form options to separate predicates

JonasVanHoof:
## Issue
 [LMB-326](https://binnenland.atlassian.net/browse/LMB-326)

## Description
Now we are defining input field options as json in the form:options predicate on our field. We want to move away from this as using separate predicates for it will make it easier to work with.

In This PR we will make both ways of adding the options to a field possible.

## Test
In the dummy app an extra field is added for each component that is using the form:options object. 
1. Use both ways separatly
2. Make use of a combination of separate predicates and the form:options as json


addon/components/rdf-input-fields/concept-scheme-selector.js
```diff
@@ -32,21 +33,28 @@ export default class RdfInputFieldsConceptSchemeSelectorComponent extends InputF
...
-    const fieldOptions = this.args.field.options;
+    const conceptSchemeUri = this.findFieldOption(
+      'conceptScheme',
+      EXT('conceptScheme')
+    );
+    const isSearchEnabled = this.findFieldOption(
+      'searchEnabled',
+      EXT('searchEnabled'),
+      'boolean'
+    );
```
JonasVanHoof, line 44: Should we make a findAllFieldOptions method that will have an array of property/predicate values?

addon/components/rdf-input-fields/input-field.js
```diff
@@ -7,10 +7,14 @@ import {
   validationResultsForField,
   validationsForFieldWithType,
 } from '@lblod/submission-form-helpers';
+import { Namespace } from 'rdflib';
 
 const MAX_LENGTH_URI = 'http://lblod.data.gift/vocabularies/forms/MaxLength';
 const SH_WARNING = SHACL('Warning');
 
+// TODO: Move EXT to the helpers namespaces.js
+export const EXT = new Namespace('http://mu.semte.ch/vocabularies/ext/');
```
JonasVanHoof, line 16: TODO

addon/components/rdf-input-fields/input-field.js
```diff
@@ -126,6 +134,51 @@ export default class InputFieldComponent extends Component {
...
+      undefined,
+      this.storeOptions.formGraph
+    );
+
+    if (predicateValue) {
+      const value = predicateValue.value;
+      const formatMap = {
+        ['string']: value,
+        ['boolean']: value == '1',
+      };
```
JonasVanHoof, line 151: Not sure on how to handle it differently  at the moment.

addon/components/rdf-input-fields/input-field.js
```diff
@@ -126,6 +134,51 @@ export default class InputFieldComponent extends Component {
...
+    if (!fieldOptions) {
+      return null;
+    }
+
+    let options = {};
+    try {
+      options = JSON.parse(fieldOptions.value);
+
+      if (options[propertyName]) {
+        return options[propertyName];
```
JonasVanHoof, line 176: Add the formatting for type also here

tests/dummy/public/test-forms/basic-fields/form.ttl
```diff
@@ -249,6 +249,17 @@ ext:conceptSchemeSelectorField a form:Field ;
 
 ext:mainFg form:hasField ext:conceptSchemeSelectorField .
 
+ext:conceptSchemeSelectorWithOptionPredicatesField a form:Field ;
+    sh:name "Select with form:options as separate predicates" ;
+    sh:order 270 ;
+    sh:path ext:conceptSchemeSelectorValue ;
+    ext:conceptScheme "http://example-concept-schemes/concept-schemes/foo-bar-baz" ;
```
JonasVanHoof, line 256: Thinking to make it possible to just pass it as prefix:predicate instead of the string. This is done like this atm as the json options handled it as a string

JonasVanHoof (commented):
Continue

addon/components/rdf-input-fields/concept-scheme-selector.js
```diff
@@ -32,19 +31,25 @@ export default class RdfInputFieldsConceptSchemeSelectorComponent extends InputF
 
   loadOptions() {
     const metaGraph = this.args.graphs.metaGraph;
-    const fieldOptions = this.args.field.options;
+    const conceptScheme = this.findFieldOption(
+      'conceptScheme',
+      EXT('conceptScheme'),
+      'node'
```
JonasVanHoof, line 37: Set from constant. RETURN_TYPE in the inputField component?

addon/components/rdf-input-fields/concept-scheme-selector.js
```diff
@@ -32,19 +31,25 @@ export default class RdfInputFieldsConceptSchemeSelectorComponent extends InputF
...
+      EXT('searchEnabled'),
+      'boolean'
+    );
 
-    if (!hasValidFieldOptions(this.args.field, ['conceptScheme'])) {
+    if (!conceptScheme) {
       return;
     }
-
-    const conceptScheme = new namedNode(fieldOptions.conceptScheme);
```
JonasVanHoof, line 41: Can be removed as we add the returnType node to the format method

addon/components/rdf-input-fields/input-field.js
```diff
@@ -126,8 +134,63 @@ export default class InputFieldComponent extends Component {
...
+    }
+  }
 }
 
 function invalidResults(validationResults) {
   return validationResults.filter((result) => !result.valid);
 }
+
+function valueToType(value, returnType) {
+  if (returnType == 'string') {
```
JonasVanHoof, line 179: Better naming + to utils method so we can add a unit test for it?

tests/dummy/public/test-forms/basic-fields/form.ttl
```diff
@@ -249,6 +250,17 @@ ext:conceptSchemeSelectorField a form:Field ;
 
 ext:mainFg form:hasField ext:conceptSchemeSelectorField .
 
+ext:conceptSchemeSelectorWithOptionPredicatesField a form:Field ;
+    sh:name "Select with form:options as separate predicates" ;
+    sh:order 270 ;
+    sh:path ext:conceptSchemeSelectorValue ;
+    ext:conceptScheme exampleConceptSchemes:foo-bar-baz ;
+    ext:searchEnabled false ;
```
JonasVanHoof, line 258: This already the example for the form:options to the separate predicates

JonasVanHoof (commented):
WIP with checkup @Rahien

addon/components/rdf-input-fields/concept-scheme-selector.js
JonasVanHoof: No just string

addon/components/rdf-input-fields/input-field.js
JonasVanHoof: To utils method and test it

addon/components/rdf-input-fields/input-field.js
```diff
@@ -126,8 +134,63 @@ export default class InputFieldComponent extends Component {
       })
     );
   }
+
+  findFieldOption(propertyName, predicate, returnType = 'string') {
```
JonasVanHoof, line 138: Change `propertyName, predicate` to one parameter and check on isNamedNode or typeof string.

+ add comment that this is done for backwards compatibility

addon/components/rdf-input-fields/alert.js
```diff
@@ -2,6 +2,7 @@ import Component from '@glimmer/component';
 
 export default class AlertComponent extends Component {
   get options() {
+    // Not changin this to findFieldOption() See input-field.js
     return this.args.field.options;
   }
```
JonasVanHoof, line 7: Here are the field.options also used. Just like int the case-number and the heading component. Should we use our new way to get those?

* ext:title
* ext:skin
* ext:level
* ext:prefix
* ext:icon
* ext:closable
* ext:size

addon/components/custom-submission-form-fields/bestuursorgaan-selector/show.js
```diff
@@ -23,10 +24,19 @@ export default class CustomSubmissionFormFieldsBestuursorgaanSelectorShowCompone
...
+
   loadOptions() {
     const metaGraph = this.args.graphs.metaGraph;
-    const fieldOptions = this.args.field.options;
-    const conceptScheme = new namedNode(fieldOptions.conceptScheme);
+    const conceptScheme = this.findFieldOption('conceptScheme', 'node');
+
+    if (!conceptScheme) {
+      return;
+    }
```
JonasVanHoof, line 39: Before there was no check if the conceptscheme was not null. This is now added. ok?

addon/components/rdf-input-fields/input-field.js
JonasVanHoof: Done

addon/components/rdf-input-fields/input-field.js
JonasVanHoof: Done by using this getter
```js
  getOptionPredicates() {
    return {
      conceptScheme: EXT('conceptScheme'),
      searchEnabled: EXT('searchEnabled'),
    };
  }

File: addon/components/rdf-input-fields/concept-scheme-selector.js (hunk 1 above)
Comment (JonasVanHoof) [reply addon/components/rdf-input-fields/concept-scheme-selector.js]: As how it is now I don't mind it having it separatly
```js
    const conceptScheme = this.findFieldOption('conceptScheme', 'node');
    const isSearchEnabled = this.findFieldOption('searchEnabled', 'boolean');
```

Windvis (commented):
I understand you want to add some helpers to make checking both the "legacy" options and the new predicates easier, but that is only needed for the existing fields. New fields (or new options for existing fields) don't need this at all right? They can simply check for the predicate they want themselves.

I fear we will start using this setup for new fields / options as well, and I don't think that's a good idea. It would also make it possible to keep using the json options, even for new options which is the opposite of what we want to achieve.

I think we should try to keep it simple. The fields that support the field.options could just check that first and manually check the new predicates afterwards if the value is undefined.

tests/dummy/public/test-forms/basic-fields/form.ttl
```diff
@@ -262,6 +302,17 @@ ext:conceptSchemeMultiSelectorField a form:Field ;
 
 ext:mainFg form:hasField ext:conceptSchemeMultiSelectorField .
 
+ext:conceptSchemeMultiSelectorWithOptionPredicatesField a form:Field ;
+    sh:name "Multiselect with form:options as separate predicates" ;
+    sh:order 280 ;
+    sh:path ext:conceptSchemeMultiSelectorValue ;
+    ext:conceptScheme exampleConceptSchemes:foo-bar-baz ;
+    ext:searchEnabled false ; # This is not even working with the json options for this component
```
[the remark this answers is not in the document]
Windvis: What doesn't work exactly? I tested it again in Firefox, and setting this to false does seem to disable the search field? There is at least a difference compared to setting this to `true` (or removing it, since true is the default).

addon/components/custom-submission-form-fields/bestuursorgaan-selector/show.js
Windvis: I think so, yes. What did it do now if the conceptScheme wasn't set? The .match would see `undefined` as a wildcard, right? So potentially it would return concepts from other conceptSchemes as well?

If so this is just a bugfix I think?

addon/components/rdf-input-fields/input-field.js
Windvis: Relevant discussion: https://github.com/lblod/submission-form-helpers/pull/62#discussion_r1457644560

So the consensus there was that we shouldn't do that.

I'm not sure we should be using `ext` prefix for the form options either. Maybe we can create a new prefix specifically for the options? Something like `field:someField formOption:someOption true`?

JonasVanHoof: Alright I'll remove the todo and when speaking of a predix for the form options itself that could be a good solution. Just like in the linked comment ext is more of a temporary prefix but thats somehting we used to give an example of the implementation here.

Should we check with someone to create a new prefix for this usecase?

addon/components/custom-submission-form-fields/bestuursorgaan-selector/show.js
JonasVanHoof: Yes indeed it would use undefined in the match what results in getting all values that have the subject and inscheme predicate.

Bug fix indeed 🆗

JonasVanHoof:
> I understand you want to add some helpers to make checking both the "legacy" options and the new predicates easier, but that is only needed for the existing fields. New fields (or new options for existing fields) don't need this at all right? They can simply check for the predicate they want themselves.
> 
> I fear we will start using this setup for new fields / options as well, and I don't think that's a good idea. It would also make it possible to keep using the json options, even for new options which is the opposite of what we want to achieve.
> 
> I think we should try to keep it simple. The fields that support the field.options could just check that first and manually check the new predicates afterwards if the value is undefined.

Im a little confused with your answer. So you want the user to use both ways for adding form:options to a field. Through the predicate form:options as string/jsonObject and throught the formOptions separate predicates. The current implementation is supporting this. Every component defines what options they need for there logic the only thing here is that the logic for getting it is in the base input class and you want to move it to a utils method. Or am I missing something? 

Moving it to a util method looks alright for me. The only thing will be that we have pass all the needed parameters instead of just accessing them from the base input field component.

So todo move the logic away from the base input to a util method?

Windvis:
Sorry, let me (try to) clarify 😄.

The idea behind this change is to add new option predicates, while still supporting the `form:options` setup for the already existing form fields. I think that is fine. I just think the implementation is more complex than it needs to be.

In my mind we simply want to do something like this:

```js
// Some field component that already supported form:options 'publicly'

//Before
let someOption = this.args.field.options.someOption;

//After
let someOption =  this.args.field.options.someOption;

if (typeof someOption === 'undefined') {
  // check if the predicate equivalent has some value, which can be as simple as doing a store.match
}
```

The whole `getOptionsPredicates` map and type conversion seems overkill to me. It _feels_ as if we want to propose this as public API that all (new) fields should start using for (new) field options in the future but I don't think that's needed.

So, I think we should try to keep it as simple as possible. If you still want to stick with the utils I wouldn't make them public API and only allow internal use, so we can refactor in the future if needed.

Windvis:
@JonasVanHoof I think the "breaking" label isn't needed, right?

Windvis:
> I think the extra if will make it more bloated but lets check 

Yes, for sure, and there are probably ways to simplify it then (hard to say for me at the moment). I just feel it's better to keep things simple and don't push this setup as the recommended way to do options.

> I'll create a new PR for that implementation so we can compare.

In any case, this is just one opinion, maybe @Rahien or @aatauil have a different one and it's fine as-is.

I would still prefer not "leaking" this into any custom components that might be out there though 😄. It would be an extra thing to support.

PR #189, opened 2024-04-17
LMB-326: Form options to separate predicates v2

JonasVanHoof:
Support form:options on a field but also form options as separate predicates e.g. formOption:conceptScheme.  

Other way of implementing https://github.com/lblod/ember-submission-form-fields/pull/188

### DIFF
--- a/addon/components/rdf-input-fields/concept-scheme-selector.js
+++ b/addon/components/rdf-input-fields/concept-scheme-selector.js
@@ -58,7 +58,19 @@ export default class RdfInputFieldsConceptSchemeSelectorComponent extends InputF
     /**
      * NOTE: Most forms are now implemented to have a default "true" behavior
      */
-    if (fieldOptions.searchEnabled !== undefined) {
+    if (!hasValidFieldOptions(this.args.field, ['searchEnabled'])) {
+      const isEnabled = this.args.formStore.any(
+        this.args.field.uri,
+        FORM_OPTION('searchEnabled'),
+        undefined,
+        this.args.graphs.formGraph
+      );
+      if (!isEnabled) {
+        return;
+      }
+
+      this.searchEnabled = isEnabled == '1' ? true : false;
+    } else {
       this.searchEnabled = fieldOptions.searchEnabled;
     }
 
--- a/addon/components/rdf-input-fields/concept-scheme-selector.js
+++ b/addon/components/rdf-input-fields/concept-scheme-selector.js
@@ -7,8 +7,12 @@ import {
   triplesForPath,
   updateSimpleFormValue,
 } from '@lblod/submission-form-helpers';
-import { namedNode } from 'rdflib';
+import { Namespace, namedNode } from 'rdflib';
 import { hasValidFieldOptions } from '../../utils/has-valid-field-options';
+
+const FORM_OPTION = new Namespace(
+  'http://mu.semte.ch/vocabularies/ext/form-option'
+);
 
 function byLabel(a, b) {
   const textA = a.label.toUpperCase();
@@ -33,12 +37,23 @@ export default class RdfInputFieldsConceptSchemeSelectorComponent extends InputF
   loadOptions() {
     const metaGraph = this.args.graphs.metaGraph;
     const fieldOptions = this.args.field.options;
+    let conceptScheme = null;
 
     if (!hasValidFieldOptions(this.args.field, ['conceptScheme'])) {
-      return;
-    }
+      const scheme = this.args.formStore.any(
+        this.args.field.uri,
+        FORM_OPTION('conceptScheme'),
+        undefined,
+        this.args.graphs.formGraph
+      );
+      if (!scheme) {
+        return;
+      }
 
-    const conceptScheme = new namedNode(fieldOptions.conceptScheme);
+      conceptScheme = scheme;
+    } else {
+      conceptScheme = new namedNode(fieldOptions.conceptScheme);
+    }
 
     /**
      * NOTE: Most forms are now implemented to have a default "true" behavior
@@ -50,6 +65,7 @@ export default class RdfInputFieldsConceptSchemeSelectorComponent extends InputF
     this.options = this.args.formStore
       .match(undefined, SKOS('inScheme'), conceptScheme, metaGraph)
       .map((t) => {
+        console.log({ t });
         const label = this.args.formStore.any(
           t.subject,
           SKOS('prefLabel'),
--- a/addon/components/rdf-input-fields/concept-scheme-selector.js
+++ b/addon/components/rdf-input-fields/concept-scheme-selector.js
@@ -77,7 +77,6 @@ export default class RdfInputFieldsConceptSchemeSelectorComponent extends InputF
     this.options = this.args.formStore
       .match(undefined, SKOS('inScheme'), conceptScheme, metaGraph)
       .map((t) => {
-        console.log({ t });
         const label = this.args.formStore.any(
           t.subject,
           SKOS('prefLabel'),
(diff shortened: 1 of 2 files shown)


tests/dummy/public/test-forms/basic-fields/form.ttl
```diff
@@ -249,6 +249,17 @@ ext:conceptSchemeSelectorField a form:Field ;
 
 ext:mainFg form:hasField ext:conceptSchemeSelectorField .
 
+ext:conceptSchemeSelectorWithFormOptionPredicatesField a form:Field ;
+    sh:name "Select with form options" ;
+    sh:order 270 ;
+    sh:path ext:conceptSchemeSelectorValue ;
+    formOption:conceptScheme exampleConceptSchemes:foo-bar-baz ;
+    formOption:searchEnabled true ;
```
JonasVanHoof, line 257: Created prefix from the ext prefix

addon/components/rdf-input-fields/concept-scheme-selector.js
```diff
@@ -33,17 +37,40 @@ export default class RdfInputFieldsConceptSchemeSelectorComponent extends InputF
...
+        this.args.field.uri,
+        FORM_OPTION('searchEnabled'),
+        undefined,
+        this.args.graphs.formGraph
+      );
+      if (!isEnabled) {
+        return;
+      }
+
+      this.searchEnabled = isEnabled == '1' ? true : false;
```
JonasVanHoof, line 72: Could convert it depending on datatype

addon/components/rdf-input-fields/concept-scheme-selector.js
```diff
@@ -33,17 +37,40 @@ export default class RdfInputFieldsConceptSchemeSelectorComponent extends InputF
...
+    } else {
       this.searchEnabled = fieldOptions.searchEnabled;
     }
```
JonasVanHoof, line 75: This is a comparison of checking it in the component itself or using it through the base input component.
```js
  getOptionPredicates() {
    return {
      conceptScheme: EXT('conceptScheme'),
    };
  }

  loadOptions() {
    const metaGraph = this.args.graphs.metaGraph;
    const conceptScheme = this.findFieldOption('conceptScheme', 'node');

    if (!conceptScheme) {
      return;
    }

File: addon/components/rdf-input-fields/concept-scheme-selector.js [hunk 4]
```diff
@@ -33,17 +37,40 @@ export default class RdfInputFieldsConceptSchemeSelectorComponent extends InputF
   loadOptions() {
     const metaGraph = this.args.graphs.metaGraph;
     const fieldOptions = this.args.field.options;
+    let conceptScheme = null;
 
     if (!hasValidFieldOptions(this.args.field, ['conceptScheme'])) {
-      return;
-    }
+      const scheme = this.args.formStore.any(
```
Rahien, line 43: if we are going this route, i would switch the order of the checks, first checking for the predicate version as that will become the default and the old one will be deprecated.

addon/components/rdf-input-fields/concept-scheme-selector.js
Rahien: there are also more components where we need to add the new way of fetching options (by predicate) right? this PR is just an example for now?

Rahien (changes requested):

addon/components/rdf-input-fields/concept-scheme-selector.js
Windvis: You could also use the `Literal.toJS` util I think: http://linkeddata.github.io/rdflib.js/doc/classes/Literal.html#toJS

addon/components/rdf-input-fields/concept-scheme-selector.js
Windvis: @Rahien Yes, it only an example so we can compare the implementations and choose which option to go for.

addon/components/rdf-input-fields/concept-scheme-selector.js
JonasVanHoof: Yes nice one works great

addon/components/rdf-input-fields/concept-scheme-selector.js
JonasVanHoof: updated it so it first checks the predicate and than the form:options if nothing was found for the predicate. Good one 👌🏻

JonasVanHoof:
@Rahien @Windvis  I updated all the input field components that use form:options. There are also components like the alert and the case-number that are also using these form:options. Should we also update them?

For the components that are updated at this moment: I added a getter for the field options. First did it inline just getting it from the store but it felt that the method was becoming to big. That's why I used the extra method :)

tests/dummy/public/test-forms/basic-fields/form.ttl
JonasVanHoof: Not sure what uri to take for the `formOptions` prefix

Ref: https://chat.semte.ch/group/form-builder?msg=Q7ETZb9rqAE569WHF

Windvis:
(FYI, I won't really have time to review this before next week (I don't work on fridays), so hopefully someone else can chime in 🙏. I think everything that doesn't increase the public API is mergeable since we can always refactor when needed.)

bfidlers:
Looks good, but I'm wondering if it wouldn't be helpful to write a helper function to parse the options? It's always pretty much the same thing you do, sometimes it is a bit different if it is a boolean or ... but should be easy enought to add this I think and would prevent a lot of similar code in future uses, plus makes it easier to use for other people.

JonasVanHoof:
Added a new page to the wiki where the possibilities are explained for the field options https://github.com/lblod/ember-submission-form-fields/wiki/Field-options

tests/dummy/app/routes/application.js
```diff
@@ -2,9 +2,9 @@ import Route from '@ember/routing/route';
 import { mockRequestsWorker } from 'dummy/mocks/worker';
 
 export default class ApplicationRoute extends Route {
-  // beforeModel() {
-  //   return mockRequestsWorker.start({
```
Windvis: What was the issue again? We should get that resolved so you don't have to keep commenting this locally 😄

Windvis (commented):
Looks good to me! 👍  I would like a definite answer about the namespace URI in the internal chat before merging though 😄.

tests/dummy/public/test-forms/basic-fields/form.ttl
```diff
@@ -106,31 +107,45 @@ ext:conceptSchemeMultiSelectCheckboxesField a form:Field ;
...
+    sh:name "Concept scheme multi select checkbox with predicate field options" ;
+    sh:order 193 ;
+    sh:path ext:conceptSchemeMultiSelectCheckboxesValue ;
+    form:validatedBy
+      [ a form:RequiredConstraint ;
+        form:grouping form:Bag ;
+        sh:resultMessage "Dit veld is verplicht."@nl;
+        sh:path ext:conceptSchemeMultiSelectCheckboxesValue] ;
+    fieldOption:conceptScheme exampleConceptSchemes:foo-bar-baz ;
+    fieldOption:orderBy qb:order ;
```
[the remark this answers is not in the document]
JonasVanHoof: Alright will  update the exising ones 👌🏻

tests/dummy/app/routes/application.js
JonasVanHoof: Haha I indeed commented this part everytime I wanted to edit something. Just ran the dummy app without commenting it and it did nothing strange! So you probably fixed it? Or mAgIc 😄

tests/dummy/public/test-forms/basic-fields/form.ttl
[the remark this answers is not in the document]
JonasVanHoof: Updated and added the form:options way of adding configuration to a field as predicated in the wiki page
image

JonasVanHoof:
We ended up choosing for the field options uri `<http://lblod.data.gift/vocabularies/field-options/>`

Windvis (approved):

PR #190, opened 2024-04-22
bugfix allow clearing of numerical-input

bfidlers:
Before clearing a numerical-input component resulted in the value being zero.
Now it is actually possible to clear to value to null, which means the triple gets deleted.


Windvis (approved):
Looks good, thanks!

Windvis:
Released as part of [v2.22.0](https://github.com/lblod/ember-submission-form-fields/blob/master/CHANGELOG.md#v2220-2024-04-23)

PR #191, opened 2024-04-24
LMB-326: update the uri to form-field-options

JonasVanHoof:
We have set the uri for field options wrong so we need to update it now as it is not used yet. 

Change the fieldOptions uri to `form-field-options`

```
@prefix fieldOption: <http://lblod.data.gift/vocabularies/form-field-options/> .
```

### DIFF
--- a/addon/utils/namespaces.js
+++ b/addon/utils/namespaces.js
@@ -1,5 +1,5 @@
 import { Namespace } from 'rdflib';
 
 export const FIELD_OPTION = new Namespace(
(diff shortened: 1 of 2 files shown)


addon/utils/namespaces.js
```diff
@@ -1,5 +1,5 @@
 import { Namespace } from 'rdflib';
 
 export const FIELD_OPTION = new Namespace(
```
JonasVanHoof, line 3: @Windvis  should we also update FIELD_OPTION to FORM_FIELD_OPTION?

PR #193, opened 2024-10-21
optionally use lazy generators

Rahien:
this allows forms to specify that they'd like to use the lazy generators that reuse the heads of the paths in the generator shapes if they exist and generate new uris for the tail ends of the paths that don't.
That way, if a generator has run and the form source data already contains generated instances, they are not generated again but rather reused.

This is completely optional and is done by setting `@lazyGenerators={{true}}` in the template

### DIFF
--- a/addon/components/rdf-form.js
+++ b/addon/components/rdf-form.js
@@ -2,6 +2,7 @@ import Component from '@glimmer/component';
 import {
   generatorsForNode,
   triplesForGenerator,
+  triplesForLazyGenerator,
 } from '@lblod/submission-form-helpers';
 import { getRootNodeForm, getTopLevelSections } from '../utils/model-factory';
 import isLast from '@lblod/ember-submission-form-fields/-private/helpers/is-last';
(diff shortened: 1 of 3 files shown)


JonasVanHoof (approved):
This is amazing! Tested it with lazyGenerators enabled and with the argument not set in the RDF form (our instance and new instance template)

The version update of the helpers that you did should not be a problem when I look at the changelog of the package. I do think you forgot to update the changelog in the helpers with the new update maybe to do something when you have some spare time 
https://github.com/lblod/submission-form-helpers/blob/master/CHANGELOG.md

Rahien:
> This is amazing! Tested it with lazyGenerators enabled and with the argument not set in the RDF form (our instance and new instance template)
> 
> The version update of the helpers that you did should not be a problem when I look at the changelog of the package. I do think you forgot to update the changelog in the helpers with the new update maybe to do something when you have some spare time https://github.com/lblod/submission-form-helpers/blob/master/CHANGELOG.md

Ah thanks! will do!

PR #194, opened 2024-12-04
Add a `form:fieldDisabled` field option

x-m-el:
This allows you to display some fields in "show" mode, while the rest of the form is editable, which can be useful to display some data without having to create custom fields for it.


Windvis:
Not sure what's up with CI. It seems to complain about this line: https://github.com/frogcat/ttl2jsonld/blob/686ae54dd13c5769750a0dd879c5551c8e1ca7ad/ttl2jsonld.js#L4617

But don't understand why that line is being executed during the build and why it only starts error'ing now :thinking:

addon/models/field.js
```diff
@@ -77,6 +83,12 @@ export default class FieldModel {
     return this.rdflibDisplayType && this.rdflibDisplayType.value;
   }
 
+  @tracked
+  rdflibDisplayShow = null;
+  get displayShow() {
+    return (this.rdflibDisplayShow && this.rdflibDisplayShow.value) === '1';
```
Windvis, line 89: I would use `Literal.toJs` so you don't have to do this `=== '1'` check.

tests/dummy/public/test-forms/basic-fields/form.ttl
```diff
@@ -158,6 +158,20 @@ ext:numericalInputField a form:Field ;
...
 
+##########################################################
+# Disabled input
+##########################################################
+ext:disabledInputField a form:Field ;
+    sh:name "Disabled input" ;
+    sh:order 223 ;
+    sh:path ext:inputValue ;
+    form:displayType displayTypes:defaultInput ;
+    form:displayShow true ;
```
Windvis, line 169: Was the predicate name of this discussed with Felix? Just double checking since it's pretty final after we merge this :smile:.

Windvis (approved):
Looks good, only a small nitpick :+1:

tests/dummy/public/test-forms/basic-fields/form.ttl
x-m-el: No, see this is a draft "POC" PR ;) 
As I indeed miss the needed context and history to make a correct decision on this.

cecemel: `form:fieldAlwaysDisabled?`

Windvis: We went for `form:fieldDisabled`

PR #195, opened 2024-12-11
Fix "true" value handling in boolean components (checkbox/switch)

x-m-el:
When creating a new entity with a form, a boolean value will be send as object `true` (which is valid ttl syntax).
Interestingly, when updating an entity (which would happen via a PUT request), the ttl will instead use `"true"^^xsd:boolean`. (as far as I could find, also valid).

When requesting the boolean value, the same syntax (`"true"^^xsd:boolean`) is used, but the boolean-aware components expect (`"1"^^xsd:boolean`). This makes it so they can handle both "true" and "1".
Fixing it here seemed the most logical.

some helper function might make sense, so it can also be used by developers creating custom components?

### DIFF
--- a/addon/components/rdf-input-fields/checkbox.js
+++ b/addon/components/rdf-input-fields/checkbox.js
@@ -9,7 +9,9 @@ export default class RdfInputFieldsCheckboxComponent extends SimpleInputFieldCom
     const matches = triplesForPath(this.storeOptions);
     if (matches.values.length > 0) {
       this.nodeValue = matches.values[0];
-      this.value = matches.values[0].value === '1'; // There is a bug in conversion from rdflib
+      // There is a bug in conversion from rdflib
+      this.value =
+        matches.values[0].value === '1' || matches.values[0].value === 'true';
     }
   }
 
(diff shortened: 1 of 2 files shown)


Windvis:
I think rdflib's `Literal.toJS` util handles this properly so we can probably use that instead.

addon/components/rdf-input-fields/checkbox.js
```diff
@@ -9,7 +9,9 @@ export default class RdfInputFieldsCheckboxComponent extends SimpleInputFieldCom
     const matches = triplesForPath(this.storeOptions);
     if (matches.values.length > 0) {
       this.nodeValue = matches.values[0];
-      this.value = matches.values[0].value === '1'; // There is a bug in conversion from rdflib
+      // There is a bug in conversion from rdflib
+      this.value =
+        matches.values[0].value === '1' || matches.values[0].value === 'true';
```
Windvis, line 14: Just tested it quickly and the following seems to work as expected.

```suggestion
      this.value = Literal.toJS(this.nodeValue);
```

And then importing the util:

```js
import { Literal } from 'rdflib';
```

Windvis:
Also, you can now rebase on master to fix the CI issues.

Windvis (approved):

PR #196, opened 2024-12-17
Allow HTML to be rendered in alert

benjay10:
Allow HTML to be rendered in alert component.

HTML should only be coming from the semantic forms and never contain user input in any way. Allowing HTML to be rendered was already done in some places, e.g. https://github.com/lblod/ember-submission-form-fields/blob/e917435aae617b72c5143f04a211d46da99129bf/addon/components/submission-form.hbs#L27

**TODO: (before or after the PR review/merge?)**

* Update the CHANGELOG
* Prepare a release

### DIFF
--- a/addon/components/rdf-input-fields/alert.hbs
+++ b/addon/components/rdf-input-fields/alert.hbs
@@ -6,5 +6,6 @@
   @skin={{this.options.skin}}
   class="au-u-margin-bottom-none"
 >
-  {{@field.label}}
+  {{! template-lint-disable no-triple-curlies}}
+  {{{@field.label}}}
 </AuAlert>


claire-lovisa (commented):
I think it's indeed OK as it's only used via the forms and no user input can mess it up. Maybe worth checking the woodpecker checks that are failing before merging?

benjay10:
I have tried looking after the warnings in the Woodpecker builds, but most of them are related to deprecations inside Appuniversum.
Some Webpack related warnings I could not fix within an hour of trying, so I am ignoring those errors for this PR as it would take longer to fix than the  work that was put in this PR. Even so, the changes in this PR could certainly not be the cause for these warnings, so I would leave fixing them for a different issue.

Windvis:
>  Some Webpack related warnings I could not fix within an hour of trying, so I am ignoring those errors for this PR as it would take longer to fix than the work that was put in this PR. Even so, the changes in this PR could certainly not be the cause for these warnings, so I would leave fixing them for a different issue.

Yea, those issues were already present on the master branch because of the changes in Appuniversum v3.7.0 (and the Embroider scenarios use floating dependencies). That issue should be fixed now on master :+1:

Windvis:
@benjay10 FYI, this should have been released as a minor instead of a patch since it's a new feature :sweat_smile:

PR #199, opened 2025-01-14
Add support for ember-concurrency v4 and power select v8

Windvis:
This allows apps to start using these newer versions as well.

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -2,12 +2,25 @@
 
 const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 
+const webpackConfig = {
+  resolve: {
+    fallback: {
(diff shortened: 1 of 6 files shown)


package-lock.json
Windvis, line 1: I needed to reroll the lock file for npm to not complain about dependency issues.

tests/dummy/public/mockServiceWorker.js
Windvis, line 1: mjs updated this when the newer version was installed.

ember-cli-build.js
```diff
@@ -2,13 +2,26 @@
 
 const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 
+const webpackConfig = {
```
Windvis, line 5: It seems ember-auto-import needs the same fix, we just didn't notice because we don't run a non-embroider scenario without a lock file.

index.js
Windvis, line 1: I copied this setup from the Appuniversum repo, since we did the same change there.

PR #218, opened 2026-01-09
Improved submission remote-data-objects list

benjay10:
**[DL-5428]**

When you open a Submission in the Toezicht module, the list of included documents is cryptic.
The aim is to make the "Links naar documenten" list of `remote-url`s more readable.

Work in progress over multiple sprints.

Also see these PR's

* `frontend-loket`: https://github.com/lblod/frontend-loket/pull/448
* `app-digitaal-loket`: to come at the end

**To test:**

See instructions on https://github.com/lblod/frontend-loket/pull/448

**To do:**

Make a release.


addon/components/rdf-input-fields/remote-urls/remote-data-object.hbs
```diff
@@ -0,0 +1,98 @@
+{{#if @model.downloadFailed}}
```
Windvis: `@model` is a vague name which we shouldn't try to extend to random component arguments. It's not obvious what `@model` is exactly without looking at what's passed in.

Naming it `@remoteDataObject` would be more clear.

addon/components/rdf-input-fields/remote-urls/show.js
```diff
@@ -61,7 +68,18 @@ export default class FormInputFieldsRemoteUrlsShowComponent extends Component {
       }
     }
 
+    for (const remoteUrl of remoteUrls) {
+      const creator = await remoteUrl.creator;
+      if (creator === 'http://lblod.data.gift/services/automatic-submission-service'
```
Windvis: For someone without the background context of how this data gets here, this code looks a bit vague.

Could you maybe try to explain in a comment what the difference is between these creator values and why the separation is needed?

addon/components/rdf-input-fields/remote-urls/show.js
```diff
@@ -61,7 +68,18 @@ export default class FormInputFieldsRemoteUrlsShowComponent extends Component {
       }
     }
 
+    for (const remoteUrl of remoteUrls) {
+      const creator = await remoteUrl.creator;
```
Windvis: creator is an attribute, so no await needed.

```suggestion
      const creator = remoteUrl.creator;
```

addon/components/rdf-input-fields/remote-urls/show.js
```diff
@@ -79,14 +97,15 @@ export default class FormInputFieldsRemoteUrlsShowComponent extends Component {
   }
 
   async loadRemoteDataObjectRecord(remoteObjectUri) {
-    const remoteUrls = await this.store.query('remote-url', {
+    const remoteUrls = await this.store.query('remote-data-object', {
```
Windvis: Could you update the mock service worker handler so the data fetch works in the test app?

It makes testing things easier if we don't have to proxy to a random project for a specific field.

Windvis (commented):
First feedback round, based purely on a look at the code. I'll have to yalc it into your loket branch, but wanted to comment some things already.

addon/components/rdf-input-fields/remote-urls/remote-data-object.hbs
Windvis: Nitpick, but I would inverse this check. I generally prefer to read the "happy state" first, before the error ones. It makes reading understanding the code easier, but that might be subjective.

No strong feelings about this though, so feel free to keep it like this as well.

addon/components/rdf-input-fields/remote-urls/show.hbs
```diff
@@ -19,42 +34,39 @@
...
-                  class="au-u-margin-left-small"
-                  download
-                >
-                  Download
-                </AuLinkExternal>
-              {{/if}}
-            </div>
-          </div>
-        </li>
+      {{#each this.sourceDocumentUrls as |remoteUrl|}}
```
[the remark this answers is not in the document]
benjay10: When the (automatic) submission worked according to plan, we should always have a source document. Normally just one in the case of an automatic one (when a vendor sends in a document), but multiple are allowed when creating the submission in Loket.

I see how this is unclear, so I'll fix the filtering of documents, together with this.

addon/components/rdf-input-fields/remote-urls/show.js
Windvis: > I didn't even know about the test app, until now. I tried running it, an I only got some testing form fields without data, until I saw that there is another route called `/form`. That seems to fail with error message `form Error: Error: NamedNode IRI "http://data.lblod.info/!DOCTYPE html" must not contain unencoded spaces.`. This is probably due to files not being served by the test setup. Did I not see documentation on this test app? How should I properly run this test app?

It should just be `npm start` like a regular app. It's indeed the /form route that you need, it wants the name of the form you want to render as a route param, but the navigatiaon links should prefill that. It should also navigate to the first form if you go to localhost:4200 (or whatever port it runs on).

https://github.com/lblod/ember-submission-form-fields/wiki/Getting-started#running-the-app

benjay10:
PR in draft, might close (not merge) later.

Scope shifted from implementing features in this addon to making custom components in Loket (and others). There will probably be a PR cleaning up components from this repo later.

PR #226, opened 2026-07-09
[DL-7363] Fix downloadAsZip to work with files containing special characters

DamonKennes:
[DL-7363]

Fix to allow files containing special characters in their titels (E.g. ^) to be downloaded as a zip.

TO TEST:
see https://github.com/lblod/frontend-toezicht-abb/pull/82 and https://github.com/lblod/frontend-worship-decisions/pull/47

### DIFF
--- a/addon/components/rdf-input-fields/files.js
+++ b/addon/components/rdf-input-fields/files.js
@@ -190,7 +190,7 @@ export default class RdfInputFieldsFilesComponent extends InputFieldComponent {
           );
         }
 
-        return response;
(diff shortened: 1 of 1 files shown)


Windvis (approved):

Windvis:
Thanks! Released as [v3.6.1](https://github.com/lblod/ember-submission-form-fields/blob/cfa00af69ebc27d57afc4ea8d82b4f9ade734df4/CHANGELOG.md#v361-2026-07-09)!

